PR #7, opened 2018-06-15
Define attributes in YAML format

beatngu13:
Probably less error prone and easier extensible. What do you think?


beatngu13:
We can also make this more general, e.g. name it `config.yaml` and add an extra point for `attributes`?

PR #16, opened 2018-07-04
Fix java.lang.ClassCastException: java.lang.Long cannot be cast to ja…

roesslerj:
…va.lang.Double

	at de.retest.web.WebElementPeer.toInt(WebElementPeer.java:116) [classes/:na]

### DIFF
--- a/src/main/java/de/retest/web/WebElementPeer.java
+++ b/src/main/java/de/retest/web/WebElementPeer.java
@@ -113,7 +113,7 @@ public class WebElementPeer {
 			return Math.toIntExact( Math.round( (Double) value ) );
 		}
 		if ( value instanceof Long ) {
-			return Math.toIntExact( Math.round( (Double) value ) );
+			return Math.toIntExact( Math.round( (Long) value ) );
 		}
 		throw new IllegalArgumentException( "Don't know how to convert a " + value.getClass() + " to int!" );
 	}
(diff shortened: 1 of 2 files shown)


src/main/java/de/retest/web/WebElementPeer.java
```diff
@@ -113,7 +113,7 @@ private int toInt( final Object value ) {
 			return Math.toIntExact( Math.round( (Double) value ) );
 		}
 		if ( value instanceof Long ) {
-			return Math.toIntExact( Math.round( (Double) value ) );
+			return Math.toIntExact( Math.round( (Long) value ) );
```
beatngu13: What I don't get: The method `toInt( Object )` is only invoked by `retrieveOutline()`. Here, the method parameter of `toInt( Object )` is always taken from the map `webData`, which is of type `Map<String, String>`. Hence, the parameter is always a string, which is why only the 2nd `if` statement (`value instanceof String`) will be triggered?

roesslerj: I can't tell you exactly in which situation that occurred, but that error was actually triggered...

roesslerj: So this is no "theoretical fix", but an actual fix to an actual error that occurred in the field...

roesslerj: Can we have any other changes as a separate PR and now just accept that fix and deploy it?

beatngu13: I see … will merge, but can you show the actual error message and/or how/where the error was?

PR #25, opened 2018-07-09
Improve showcase

roesslerj:


README.md
```diff
@@ -43,7 +43,7 @@ Download recheck-web [here on GitHub](https://github.com/retest/recheck-web/rele
 </dependency>
 ```
 
-Then replace the assertions in your Selenium test. An example test could look like so:
+Then replace the assertions in your Selenium test. An example test could look like so ([simplified](https://github.com/retest/recheck-web/blob/master/src/test/java/de/retest/web/ShowcaseIntTest.java)):
```
beatngu13: Link is broken, now named `ShowcaseIT`.

PR #44, opened 2018-08-02
Feature/normalize data

roesslerj:


src/main/java/de/retest/web/WebElementPeer.java
```diff
@@ -100,14 +100,21 @@ public OutlineAttribute retrieveOutline() {
 	protected MutableAttributes retrieveStateAttributes() {
 		final MutableAttributes state = new MutableAttributes();
 		for ( final String attribute : AttributesProvider.getInstance().getAttributes() ) {
-			final String attributeValue = webData.get( attribute );
+			final String attributeValue = normalize( webData.get( attribute ) );
```
diba1013: Shouldn't we do this to all attributes (i.e inside the constructor)?

src/test/java/de/retest/web/WebElementPeerTest.java
```diff
@@ -44,4 +44,9 @@ void filled_web_data_should_result_in_outline_with_corresponding_values() throws
 		assertThat( rectangle.getHeight() ).isEqualTo( 4.0 );
 	}
 
+	@Test
+	void normalize_should_remove_apostrophe() {
```
diba1013: Theoretical what if `""`, `"\""`, `"    \""`, `"\"   "`.

src/main/java/de/retest/web/WebElementPeer.java
roesslerj: Not so sure about this. I don't know how future Attributes will be implemented and for which technologies. Would refrain from generalizing too much...

src/test/java/de/retest/web/WebElementPeerTest.java
roesslerj: Blanks should be trimmed. But the other cases ... would wait for them to occur in practice, to have a better understanding.

PR #58, opened 2018-08-24
Add gif that shows how to replace asserts with "re.check()"

roesslerj:


beatngu13:
As mentioned in Slack, we should a) increase the quality a bit and b) slow down the replay speed.

roesslerj:
Done :-)

PR #65, opened 2018-09-19
Fix method isDefault

kema91:
- Fix method isDefault
- Add some tests to check the default values correctly
- Delete the unneeded defaults from the retest.xml

With the fix it should be possible that no default values are written to the retest.xml

### DIFF
--- a/src/main/java/de/retest/web/DefaultValuesProvider.java
+++ b/src/main/java/de/retest/web/DefaultValuesProvider.java
@@ -8,6 +8,8 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
 
+import org.apache.commons.lang3.StringUtils;
+
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
@@ -74,8 +76,8 @@ public class DefaultValuesProvider implements DefaultValueFinder {
 
 	public boolean isDefault( final String tag, final String attribute, final String attributeValue ) {
 		final String defaultValue = getDefaultValue( tag, attribute );
-		if ( defaultValue != null ) {
-			return defaultValue.equalsIgnoreCase( attributeValue );
+		if ( StringUtils.equalsIgnoreCase( defaultValue, attributeValue ) ) {
+			return true;
 		}
 		if ( attributeValue == null || attributeValue.trim().isEmpty() ) {
 			return true;
--- a/src/main/java/de/retest/web/DefaultValuesProvider.java
+++ b/src/main/java/de/retest/web/DefaultValuesProvider.java
@@ -8,8 +8,6 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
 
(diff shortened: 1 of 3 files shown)


src/test/java/de/retest/web/DefaultValuesProviderTest.java
```diff
@@ -45,4 +45,18 @@ void getDefaultValue_should_prefer_specific_over_general_values() {
 		assertThat( cut.getDefaultValue( "body", "margin-top" ) ).isEqualTo( "8px" );
 		assertThat( cut.getDefaultValue( "boody", "margin-top" ) ).isEqualTo( "0px" );
 	}
+
+	@Test
+	void isDefault_handled_two_null_values() {
```
beatngu13: Doesn't the test tests more than "two null values"? It also checks if stuff like `auto`, `normal`, and `0px` is recognized as a default, besides the defaults provided in `defaults.yml`.

src/main/java/de/retest/web/DefaultValuesProvider.java
```diff
@@ -74,8 +76,8 @@ public Serializable getDefaultValue( final IdentifyingAttributes comp, final Str
 
 	public boolean isDefault( final String tag, final String attribute, final String attributeValue ) {
 		final String defaultValue = getDefaultValue( tag, attribute );
-		if ( defaultValue != null ) {
-			return defaultValue.equalsIgnoreCase( attributeValue );
+		if ( StringUtils.equalsIgnoreCase( defaultValue, attributeValue ) ) {
```
beatngu13: How about moving it behind the `null` check for `attributeValue` and then to:

```java
if ( attributeValue.equalsIgnoreCase( defaultValue ) ) { ... }
```

kema91: do you mean something like that: 


`if ( attributeValue == null || attributeValue.trim().isEmpty() ) {`
    `return true;`
`}`

`if ( attributeValue.equalsIgnoreCase( defaultValue ) {`
    ` return true;`
`}`

src/test/java/de/retest/web/DefaultValuesProviderTest.java
kema91: I'll change it

beatngu13: I'll give you a call. 😉

src/main/java/de/retest/web/DefaultValuesProvider.java
```diff
@@ -74,12 +74,12 @@ public Serializable getDefaultValue( final IdentifyingAttributes comp, final Str
 
 	public boolean isDefault( final String tag, final String attribute, final String attributeValue ) {
 		final String defaultValue = getDefaultValue( tag, attribute );
-		if ( defaultValue != null ) {
-			return defaultValue.equalsIgnoreCase( attributeValue );
-		}
 		if ( attributeValue == null || attributeValue.trim().isEmpty() ) {
 			return true;
 		}
+		if ( attributeValue.equalsIgnoreCase( defaultValue ) ) {
```
roesslerj: Damn, I overlooked that. We want to be able to say, that null is not the default value...

PR #68, opened 2018-09-20
Feature/add retest workspace

kema91:
Polish recheck web (Make it to checkout and it works)
- Now it is possible to start retest from the IDE and use the function "report review" 
- Unfortunately it does not work with the current retest version 3.0


kema91:
rebaser helloooo

PR #69, opened 2018-09-20
Feature/better defaults

roesslerj:


src/main/java/de/retest/web/DefaultValuesProvider.java
```diff
@@ -83,10 +83,11 @@ public Serializable getDefaultValue( final IdentifyingAttributes comp, final Str
 	}
 
 	public boolean isDefault( final String tag, final String attribute, final String attributeValue ) {
-		if ( attributeValue == null || attributeValue.trim().isEmpty() ) {
+		final String defaultValue = getDefaultValue( tag, attribute );
```
beatngu13: >If we define a default value, than null or "" is not accepted as default

What exactly do you mean by that? In [our test](https://github.com/retest/recheck-web/blob/master/src/test/java/de/retest/web/DefaultValuesProviderTest.java#L21) we also check for `null` and `""`, which is accepted as a default. If you mean something different, could you add a test?

src/main/resources/defaults.yaml
```diff
@@ -171,15 +171,22 @@ all:
   - z-index: auto
 
 a:
+  - border-image-width: 1
```
beatngu13: This occurs multiple times. Is there a way to define it only once? I'm not that familiar with the current `defaults.yaml`.

PR #70, opened 2018-09-22
Update dependencies

beatngu13:


beatngu13:
Regarding 6430f5e: That's probably something we should fix before the release of retest v3?!

beatngu13:
rebazer?

PR #73, opened 2018-10-04
Feature/default over null

kema91:
If we have a default value then null or empty String should not be accepted.
E.g. if we have a link, then the default value for text-decoration is "underline". If instead we get `null` or `""`, then this is not the default and should be recorded as such.

```
assertThat( cut.isDefault( "a", "text-decoration", "" ) ).isFalse();
assertThat( cut.isDefault( "a", "text-decoration", null ) ).isFalse();
```


kema91:
merge?

PR #75, opened 2018-10-15
Feature/ret 1316 more drivers

kema91:
- Prepare travis to also provide Firefox driver aka geckodriver
- A factory with which all drivers can be initialized
- We are now able to test every example website with all drivers

- Currently we are testing with firefox and chrome
- Every driver needs his own Golden Master


beatngu13:
@roesslerj done with clean up, you can start your review now.

roesslerj:
I will approve this PR. But next task is to find a way to reduce the amount of "duplication" in the `SimplePageIT` for Firefox...

PR #92, opened 2018-11-05
Add logback as test dependency, so we have logging during dev...

roesslerj:


pom.xml
```diff
@@ -200,6 +200,19 @@
 			<scope>test</scope>
 		</dependency>
 
+		<dependency>
+			<groupId>ch.qos.logback</groupId>
+			<artifactId>logback-core</artifactId>
+			<version>1.2.3</version>
```
beatngu13: Would you mind adding something like `logback.version` to `properties` and then use the variable here and below (like `junit.jupiter.version`)?

PR #96, opened 2018-11-08
Add invisible parents

kema91:
We need a parent element to add a child

### DIFF
--- a/src/main/java/de/retest/web/RecheckSeleniumAdapter.java
+++ b/src/main/java/de/retest/web/RecheckSeleniumAdapter.java
@@ -106,8 +106,10 @@ public class RecheckSeleniumAdapter implements RecheckAdapter {
 				peer = root;
 			} else {
 				peer = new WebElementPeer( defaultValuesProvider, webData, path );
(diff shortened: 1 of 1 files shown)


src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -106,8 +106,10 @@ public RootElement convertToPeers( final Map<String, Map<String, Object>> data,
 				peer = root;
 			} else {
 				peer = new WebElementPeer( defaultValuesProvider, webData, path );
-				final WebElementPeer parent = converted.get( parentPath );
-				assert parent != null : "We sorted the map, parent should already be there.";
+				WebElementPeer parent = converted.get( parentPath );
+				if ( parent == null ) {
+					parent = new WebElementPeer( defaultValuesProvider, webData, parentPath );
```
roesslerj: This is wrong: You create the parent with the child data... you need to get the webData for the parent, put the parent into `converted` and add this child to the parent...

rebazer:

PR #97, opened 2018-11-08
Feature/fix missing parent

roesslerj:


kema91:
the change fixes the bug and the amazon integration test is now running

src/test/java/de/retest/web/PeerConverterTest.java
```diff
@@ -16,7 +16,7 @@
 import de.retest.ui.diff.RootElementDifference;
 import de.retest.ui.diff.RootElementDifferenceFinder;
 
-class RecheckSeleniumAdapterTest {
+class PeerConverterTest {
```
beatngu13: Although it already has been merged: `convertToPeers_should_result_in_valid_tree()` is a test that belong to `RecheckSeleniumAdapter`, i.e., a separate test. You probably also don't need the `setUp()` in `PeerConverterTest`...

PR #108, opened 2018-11-16
Feature/fix signed deploy

modulo11:


.travis.settings.xml
```diff
@@ -6,4 +6,15 @@
 			<password>${env.OSSRH_PASSWORD}</password>
 		</server>
 	</servers>
+	<profiles>
```
beatngu13: Isn't it a bit confusing to use two profiles (one here and one in the `pom.xml`)? GPG is only needed during deployment, so `activateByDefault` is also a bit misleading. What do you think?

PR #109, opened 2018-11-16
Feature/resize screenshots

kema91:


src/main/java/de/retest/web/ScreenshotProvider.java
```diff
@@ -12,20 +18,54 @@
...
+		final Image tmp = image.getScaledInstance( width, height, Image.SCALE_SMOOTH );
+		final BufferedImage resized = new BufferedImage( width, height, BufferedImage.TYPE_INT_ARGB );
+		final Graphics2D graphics2D = resized.createGraphics();
+		graphics2D.drawImage( tmp, 0, 0, null );
+		graphics2D.dispose();
+		return resized;
+	}
+
+	public static boolean isRetinaDisplay() {
+		if ( SystemUtils.IS_OS_MAC ) {
```
roesslerj: This also can be with Windows... just remove the if.

src/main/java/de/retest/web/ScreenshotProvider.java
```diff
@@ -12,20 +18,54 @@
...
+			final GraphicsEnvironment environment = GraphicsEnvironment.getLocalGraphicsEnvironment();
+			final GraphicsDevice device = environment.getDefaultScreenDevice();
+			try {
```
roesslerj: I would extract this into a separate method `extractScale` or similar.

src/main/java/de/retest/web/ScreenshotProvider.java
```diff
@@ -12,20 +18,54 @@
...
+	private static BufferedImage resize( final BufferedImage image, final int width, final int height ) {
+		final Image tmp = image.getScaledInstance( width, height, Image.SCALE_SMOOTH );
+		final BufferedImage resized = new BufferedImage( width, height, BufferedImage.TYPE_INT_ARGB );
+		final Graphics2D graphics2D = resized.createGraphics();
+		graphics2D.drawImage( tmp, 0, 0, null );
+		graphics2D.dispose();
+		return resized;
+	}
+
+	public static boolean isRetinaDisplay() {
```
roesslerj: As mentioned, this code belongs to retest-model. So add it there, and put a comment here so this method can be removed, once the change "trickles" here.

src/main/java/de/retest/web/ScreenshotProvider.java
```diff
@@ -12,20 +18,54 @@
...
+		if ( SystemUtils.IS_OS_MAC ) {
+			final GraphicsEnvironment environment = GraphicsEnvironment.getLocalGraphicsEnvironment();
+			final GraphicsDevice device = environment.getDefaultScreenDevice();
+			try {
+				final Field field = device.getClass().getDeclaredField( "scale" );
+				if ( field != null ) {
+					field.setAccessible( true );
+					final Object scale = field.get( device );
+					if ( ((Integer) scale).intValue() == RETINA_SCALE_FACTOR ) {
```
roesslerj: Shouldn't it just return the scale factor? That way we would not need to change the code, once we have a scale factor of 4 or higher...

src/main/java/de/retest/web/ScreenshotProvider.java
```diff
@@ -12,20 +18,54 @@
...
 
 	public static BufferedImage shootFullPage( final WebDriver driver ) {
+		final BufferedImage image;
 		if ( driver instanceof ChromeDriver ) {
-			return Shutterbug
+			image = Shutterbug
 					.shootPage( driver, ScrollStrategy.WHOLE_PAGE_CHROME, SCROLL_TIMEOUT, USE_DEVICE_PIXEL_RATIO )
 					.getImage();
+			if ( isRetinaDisplay() ) {
+				return resize( image, image.getWidth() / RETINA_SCALE_FACTOR, image.getHeight() / RETINA_SCALE_FACTOR );
```
roesslerj: As said, return the scale factor instead of using a fixed one.

src/main/java/de/retest/web/ScreenshotProvider.java
```diff
@@ -12,20 +18,54 @@
...
+			try {
+				final Field field = device.getClass().getDeclaredField( "scale" );
+				if ( field != null ) {
+					field.setAccessible( true );
+					final Object scale = field.get( device );
+					if ( ((Integer) scale).intValue() == RETINA_SCALE_FACTOR ) {
+						return true;
+					}
+				}
+			} catch ( final Exception ignore ) {}
```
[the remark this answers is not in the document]
roesslerj: At least we should log the exception here.

PR #133, opened 2018-12-19
Remove redundant dependency

oezelenes:


pom.xml
```diff
@@ -86,12 +86,6 @@
 			<version>0.6.0</version>
 		</dependency>
 
-		<dependency>
```
beatngu13: AFAIK this dependency is being used, right? Generally, I think we should declare dependencies explicitly and not rely on transitivity if the corresponding project uses them directly. The version chaos will go away as soon as we have a proper parent POM.

Issue #145, opened 2019-01-30
Link target not checked


beatngu13:
When changing the target of a link, for instance

```html
<li><span><a href="https://www.retest.de/en/for-testers/index.html">For Testers</a></span>
```

to

```html
<li><span><a href="SOME BROKEN URL">For Testers</a></span>
```

then the change is not detected. Consequently, we should also check `href`.

beatngu13:
See PR #152.

PR #151, opened 2019-02-01
Use all known attributes for state

### LINKED ISSUE #147: Adding id and name to div leads to no differences
Adding an `id` and a `name` to a `div` like here

https://github.com/retest/recheck-web/blob/master/src/test/resources/pages/showcase/retest.html#L192

Leads to no differences, although we are supposed to track these changes?

modulo11:
Closes #147.


src/main/java/de/retest/web/WebElementPeer.java
```diff
@@ -89,7 +89,7 @@ public Integer extractSuffix() {
 
 	protected MutableAttributes retrieveStateAttributes() {
 		final MutableAttributes state = new MutableAttributes();
-		for ( final String attribute : AttributesProvider.getInstance().getAttributes() ) {
+		for ( final String attribute : AttributesProvider.getInstance().getJoinedAttributes() ) {
```
roesslerj: This still only checks known attributes. Per default, we want _all_ (even unexpected or like for HTML5 user-defined) attributes... This then also addresses https://github.com/retest/recheck-web/issues/145

PR #152, opened 2019-02-01
Track all HTML attributes

### LINKED ISSUE #147: Adding id and name to div leads to no differences

modulo11:
We should consider filtering some of them at a later time. This should at least close #147 and #145.


beatngu13:
FYI: If we want to automatically close issues via pull request, we need [corresponding keywords](https://help.github.com/articles/closing-issues-using-keywords/) such as close *before every issue*, i.e., close #147 and close #145.

PR #154, opened 2019-02-05
Feature/fix attributes

oezelenes:


src/main/java/de/retest/web/WebElementPeer.java
```diff
@@ -89,10 +90,9 @@ public Integer extractSuffix() {
 
 	protected MutableAttributes retrieveStateAttributes() {
 		final MutableAttributes state = new MutableAttributes();
-
 		for ( final String key : webData.getKeys() ) {
 			final String attributeValue = webData.getAsString( key );
-			if ( attributeValue != null ) {
+			if ( attributeValue != null && !isIdentifyingAttributesKey( key ) ) {
```
beatngu13: Does that mean we exclude identifying attributes from the actual state?

src/main/java/de/retest/web/AttributesConfig.java
```diff
@@ -17,6 +19,13 @@
 	public static final String WIDTH = "width";
 	public static final String HEIGHT = "height";
 
+	public static final String TAG_NAME = "tagName";
+
+	public static final String TEXT = "text";
+
+	private static final HashSet<String> identifyingAttributeKeys = new HashSet<>( Arrays.asList( ABSOLUTE_HEIGHT,
```
beatngu13: Left-hand side should be `Set`.

src/main/java/de/retest/web/AttributesConfig.java
```diff
@@ -35,4 +44,8 @@ public AttributesConfig( @JsonProperty( "attributes" ) final List<String> attrib
 		return identifyingAttributes;
 	}
 
+	static boolean isIdentifyingAttributesKey( final String key ) {
```
beatngu13:
```suggestion
	static boolean isIdentifyingAttribute( final String key ) {
```

src/main/java/de/retest/web/AttributesConfig.java
beatngu13: Plus: it's somewhat odd to have two identifying attribute sets/lists. The problem is actually the underlying YAML, we should probably change.

src/main/java/de/retest/web/AttributesConfig.java
```diff
@@ -17,6 +19,13 @@
 	public static final String WIDTH = "width";
 	public static final String HEIGHT = "height";
 
+	public static final String TAG_NAME = "tagName";
```
beatngu13: Isn't tag name = type?

src/main/java/de/retest/web/WebElementPeer.java
oezelenes: No, the attribute name "state" is a bit misleading there I guess. The method is used to retrieve the attributes for an element. So we just exclude already existing IdentifyingAttributes content.

src/main/java/de/retest/web/AttributesConfig.java
oezelenes: We call it type but actaully use tagName to extract it from the data.

beatngu13:
Did some refactoring trying to make clear what we do with `attributes.yaml` and how it is used in `getAllElementsByPath.js`. We can probably also use `AttributesUtil` to bundle filter methods and stuff. What do you think @oezelenes @modulo11?

PR #160, opened 2019-02-13
Feature/fix html foramtting

oezelenes:


src/main/java/de/retest/web/WebData.java
```diff
@@ -27,7 +27,8 @@ public String getAsString( final String key ) {
 		if ( result == null ) {
 			return null;
 		}
-		return normalize( String.valueOf( result ) );
+		final String value = String.valueOf( result );
+		return key.equals( AttributesConfig.TEXT ) ? value : normalize( value );
```
beatngu13: Does it make sense to have tests for both cases?

src/main/java/de/retest/web/util/TextAttributeUtil.java
```diff
@@ -0,0 +1,22 @@
...
+
+import static de.retest.web.AttributesConfig.TEXT;
+
+import de.retest.recheck.ui.descriptors.TextAttribute;
+
+public class TextAttributeUtil {
+
+	private TextAttributeUtil() {}
+
+	public static final String PRE_ELEMENT = "pre";
```
beatngu13: Does it make sense to use someting like `/pre`, `pre/`, or `/pre` so we don't accidentally apply the same logic when something match `*pre*`?

src/main/java/de/retest/web/util/TextAttributeUtil.java
```diff
@@ -0,0 +1,22 @@
...
+
+	public static TextAttribute createTextAttribute( final String path, final String text ) {
+		return path.toLowerCase().contains( PRE_ELEMENT ) ? new TextAttribute( TEXT, text )
+				: new TextAttribute( TEXT, cleanString( text ) );
+	}
+
+	private static String cleanString( final String text ) {
```
beatngu13: How about `reduceConsecutiveWhitespaces` or something like this?

PR #162, opened 2019-02-20
Update to latest version

roesslerj:
Who approves first? :-D


beatngu13:
Current is 0.11.0, but I doubt it's already on Central…

PR #185, opened 2019-03-12
Feature/improve ignores

roesslerj:


src/test/java/de/retest/web/it/RecheckIgnoreJsIT.java
```diff
@@ -0,0 +1,66 @@
...
+import java.nio.file.Paths;
+
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+import de.retest.recheck.ignore.JSShouldIgnoreImpl;
+import de.retest.recheck.ui.descriptors.Element;
+import de.retest.recheck.ui.diff.AttributeDifference;
+
+public class RecheckIgnoreJsIT {
```
beatngu13: `public` can be removed.

.retest/recheck.ignore.js
```diff
@@ -0,0 +1,46 @@
...
+		diff.actual = "";
+	}
+	if (contains(['absolute-x', 'absolute-y', 'absolute-width', 'absolute-height', 'x', 'y', 'width', 'height'], diff.key)) {
+		return (Math.abs(diff.expected.replace("px", "") - diff.actual.replace("px", "")) <= allowedPxDiff);
+	}
+	if (diff.expected.indexOf("px") !== -1) {
+		return (Math.abs(diff.expected.replace("px", "") - diff.actual.replace("px", "")) <= allowedPxDiff);
+	}
+	if (diff.key == "font-family") {
+	    for (var i = 0; i < fontFamilies.length; i++) {
```
beatngu13: This uses spaces for indentation, whereas other lines use tabs. Could you use something like https://beautifier.io/?

.retest/recheck.ignore
```diff
@@ -0,0 +1,6 @@
+# ignore non-visible attributes
```
beatngu13: Both ignore files are simply improvements for our own recheck-web tests. If they work out, we can also think about automatically creating them in case no ignore files can be found.

.retest/recheck.ignore.js
```diff
@@ -0,0 +1,46 @@
+var allowedPxDiff = 5;
```
beatngu13: I'm not that familiar with JS, but aren't these [constants](https://stackoverflow.com/a/130399)? So if we don't want to rely on ES2015, shouldn't this be something like `ALLOWED_PIXEL_DIFF`? And the other variables/constants?

.retest/recheck.ignore.js
```diff
@@ -0,0 +1,46 @@
...
+var baseUrl = /http[s]?:\/\/[\w.:\d\-]*/;
+var fontFamilies = [["system-ui", "Arial"], ["-apple-system", "sans-serif"]];
+
+function contains(array, key) {
```
beatngu13: And again: I'm not that familiar with JS, but … 😊 Can't we use [`Array.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)? According to [MDN](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects), these objects and methods are always available in any JS environment? If this is true, maybe we can simplify more code?

.retest/recheck.ignore.js
```diff
@@ -0,0 +1,46 @@
+var allowedPxDiff = 5;
+var baseUrl = /http[s]?:\/\/[\w.:\d\-]*/;
+var fontFamilies = [["system-ui", "Arial"], ["-apple-system", "sans-serif"]];
```
beatngu13: Is the leading hyphen in `"-apple-system"` intended?

.retest/recheck.ignore
roesslerj: I would give them as examples, but like the .gitignore is not filled with e.g. `*/bin`, I would not fill those, but instead point to examples in the comment.

.retest/recheck.ignore.js
roesslerj: Yes, this is how it shows up at my system...

src/test/java/de/retest/web/it/RecheckIgnoreJsIT.java
roesslerj: Done

.retest/recheck.ignore.js
roesslerj: Now used Eclipse autoformat...

.retest/recheck.ignore
beatngu13: Indeed, good idea!

.retest/recheck.ignore.js
roesslerj: Maybe nashorn doesn't have it? `["absolute-x", "absolute-y", "absolute-width", "absolute-height", "x", "y", "width", "height"].includes is not a function in <eval> at line number 28
22:33:07.798 [main] INFO de.retest.recheck.ignore.JSShouldIgnoreImpl - Reading JS ignore rules file from .retest/recheck.ignore.js.`

.retest/recheck.ignore.js
roesslerj: Renamed...

PR #190, opened 2019-03-15
Feature/fix hidden div

roesslerj:


src/test/java/de/retest/web/it/EmptyDivTestIT.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.openqa.selenium.WebDriver;
+
+import de.retest.recheck.Recheck;
+import de.retest.recheck.RecheckImpl;
+
+class EmptyDivTestIT {
```
beatngu13: Instead of yet another IT on top of Selenium, does a unit test make sense? As I see, only `WebData` has been changed? At least it looks like `isShown()` can be easily tested.

src/test/java/de/retest/web/it/EmptyDivTestIT.java
```diff
@@ -0,0 +1,46 @@
...
+	WebDriver driver;
+	Recheck re;
+
+	@BeforeEach
+	void setUp() {
+		re = new RecheckImpl();
+	}
+
+	@ParameterizedTest
+	@MethodSource( "de.retest.web.testutils.WebDriverFactory#drivers" )
```
beatngu13: If we keep the IT: Doesn't a single driver suffice?

src/test/java/de/retest/web/it/EmptyDivTestIT.java
```diff
@@ -0,0 +1,46 @@
...
+	public void testCenter( final WebDriver driver ) throws Exception {
```
beatngu13: `public` can be removed from all methods.

src/test/java/de/retest/web/WebDataTest.java
```diff
@@ -3,7 +3,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
-import java.awt.Rectangle;
+import java.awt.*;
```
beatngu13: I guess this wasn't intended? (AFAIK we don't format `*` imports.)

PR #195, opened 2019-03-29
The problem - some attribute changes show up twice no longer occurs

ivonx:
I tested different sites and html files and i saw, that the attribute changes are shown twice when we only test the "retest.de" site. Actually the attributes are the same and one of them is not showed in the attribute field.
zweimalAngezeigt
theSameAttribute
With the filter in the WebElementPeer.java class  i fixed that and the attribute occurs only one time:
AttributeOnce
done

### DIFF
--- a/src/main/java/de/retest/web/WebElementPeer.java
+++ b/src/main/java/de/retest/web/WebElementPeer.java
@@ -93,6 +93,7 @@ public class WebElementPeer {
 		webData.getKeys().stream() //
 				.filter( Objects::nonNull ) //
 				.filter( AttributesUtil::isStateAttribute ) //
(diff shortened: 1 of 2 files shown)


src/test/java/de/retest/web/it/websites/ReTestTest.java
```diff
@@ -0,0 +1,64 @@
...
+		// Use the default implementation.
+		re = new RecheckImpl();
+	}
+
+	@Test
+	public void index() throws Exception {
+		// Set the file name of the Golden Master.
+		re.startTest( "ReTest" );
+
+		driver.get( "https://retest.de/" );
```
roesslerj: Problem here is that this creates an "outside" dependency. When we change the retest website, the build (then master!) starts to fail...
So either add lokal HTML files or do not add this test, as this is reflected in existing tests already.

PR #201, opened 2019-04-09
Improve attributes.yaml

beatngu13:
Previously, the `attributes.yaml` file contained two *lists*:

```yaml
cssAttribtutes:
  - align-content
  - …
htmlAttributes:
  - class
  - …
```

`cssAttribtutes` has been used for an element's state, whereas `htmlAttributes` contained additional, web-specific identifying attributes.

We actually want `attributes.yaml` to be used (internally) to define what CSS and HTML attributes will be part of the state. It is now possible to define *sets* for both. These sets can also be empty. In the case of `htmlAttributes`, the value `all` is also allowed, which then uses all HTML attributes for an element's state.

The default behavior (only selected CSS attributes but all HTML attributes) doesn't change with this commit, only the way `attributes.yaml` is going to be used.


src/main/java/de/retest/web/AttributesConfigDeserializer.java
```diff
@@ -0,0 +1,54 @@
...
+	private List<String> toCssAttributesList( final JsonNode cssAttributesNode ) {
+		if ( cssAttributesNode.isTextual() ) {
+			throw new IllegalArgumentException(
+					"CSS attributes can only be a list of selected attributes or empty ('all' not supported)." );
+		}
+		return toList( cssAttributesNode );
+	}
+
+	private List<String> toHtmlAttributesList( final JsonNode htmlAttributesNode ) {
+		if ( htmlAttributesNode.isTextual() ) {
```
diba1013: Do we want to check, if the node is `"all"`?

src/main/java/de/retest/web/AttributesConfigDeserializer.java
```diff
@@ -0,0 +1,58 @@
...
+		return toSet( htmlAttributesNode );
+	}
+
+	private Set<String> toSet( final JsonNode node ) {
+		if ( node.isNull() ) {
+			return Collections.emptySet();
+		}
+		final Spliterator<String> spliterator =
+				Spliterators.spliteratorUnknownSize( node.fieldNames(), Spliterator.NONNULL );
+		return StreamSupport.stream( spliterator, false ) //
```
diba1013: This should be an `ArrayNode`. Is there some methods to retrieve it as `List`? If not, can't you use the `parser` or `context` to deserialize it as `List<String>`?

src/main/resources/attributes.yaml
```diff
@@ -1,187 +1,184 @@
...
-  - unicode-bidi
-  - user-select
-  - vertical-align
-  - visibility
-  - white-space
-  - word-break
-  - word-spacing
-  - word-wrap
-  - z-index
+  ? align-content
```
diba1013: Is this `?` part of https://github.com/retest/recheck-web/pull/201/files#diff-2f56b6bae935438614bd6475920f25d8R54

src/main/java/de/retest/web/AttributesConfigDeserializer.java
```diff
@@ -0,0 +1,58 @@
...
+				.collect( Collectors.toCollection( HashSet::new ) );
```
diba1013: If you must keep it:
```suggestion
				.collect( Collectors.toSet() );
```

src/main/java/de/retest/web/AttributesConfigDeserializer.java
beatngu13: If "all" is selected, the node is textual. So this is essentially the check for it. But we could extract a method like:

```java
private boolean isAll( JsonNode node ) {
    return node.isTextual();
}
```

We could also think about more checks to see if the text is actually "all", but I don't think it is necessary at this stage.

src/main/java/de/retest/web/AttributesConfigDeserializer.java
beatngu13: In YAML, sets are maps with `null` values, which is why this is an `ObjectNode` and we have to use its field names. However, I think making the set a list and only using a set internally may be better.

src/main/java/de/retest/web/AttributesConfigDeserializer.java
beatngu13: `toSet()` [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toSet--):

>There are no guarantees on the type, mutability, serializability, or thread-safety of the `Set` returned; if more control over the returned `Set` is required, use `toCollection(Supplier)`.

Since we want fast access for contains checks, I think this is the right approach.

src/main/resources/attributes.yaml
beatngu13: As I said, that is the syntax for YAML sets.

PR #202, opened 2019-04-11
Feature/autocheckingdriver

roesslerj:


diba1013:
A strange vision again. Why do we need this?

roesslerj:
> A strange vision again. Why do we need this?

This is the essence of recheck. You don't need to specify anything. You actually don't even need to tell recheck to check anything. This is how you can use the AI. And this also allows you to combine it with record/replay easily. Remember retest for swing? This is one of it's core features...

beatngu13:
As general note: I think this should be a draft and presented to the team as part of the innovations & ideas meeting.

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
```diff
@@ -0,0 +1,93 @@
...
+	public List<WebElement> findElements( final By by ) {
+		return super.findElements( by ).stream().map( element -> new WebElementWrapper( element, this ) )
+				.collect( Collectors.toList() );
+	}
+
+	public void check( final String currentStep ) {
+		re.check( this, makeUnique( currentStep ) );
+	}
+
+	private int counter = 0;
```
beatngu13: What does this "count"? Maybe a better name like `idCounter`?

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	@Override
+	public WebElement findElement( final By by ) {
+		return new WebElementWrapper( super.findElement( by ), this );
+	}
+
+	@Override
+	public List<WebElement> findElements( final By by ) {
+		return super.findElements( by ).stream().map( element -> new WebElementWrapper( element, this ) )
```
beatngu13: Should be formatted like

```java
super.findElements( by ).stream() //
        .map( element -> new WebElementWrapper( element, this ) ) //
        .collect( Collectors.toList() );
```

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.remote.RemoteWebDriver;
+
+import de.retest.web.RecheckWebImpl;
+
+public class AutocheckingRecheckDriver extends RecheckDriver {
```
beatngu13: AFAIK the "official" mechanism for this would be to use `EventFiringWebDriver` ([example](https://seleniumeasy.com/selenium-tutorials/webdriver-event-listener-example/)). But since we already adopted the approach from `RecheckDriver`, I think it doesn't make sense to switch. (Honestly, I also don't know which is "better".)

beatngu13: Aren't "auto" and "checking" two words, i.e., `AutoCheckingRecheckDriver`?

src/main/java/de/retest/web/selenium/WebElementWrapper.java
```diff
@@ -0,0 +1,131 @@
...
+import org.openqa.selenium.WebElement;
+
+public class WebElementWrapper implements WebElement {
+
+	private final WebElement delegate;
+	private final AutocheckingRecheckDriver driver;
+
+	public WebElementWrapper( final WebElement delegate, final AutocheckingRecheckDriver driver ) {
+		if ( delegate instanceof WebElementWrapper ) {
+			throw new IllegalStateException( "Wrapping doubly!" );
```
beatngu13: Maybe `IllegalArgumentException` with `"Cannot wrap WebElementWrapper inside WebElementWrapper."`?

src/main/java/de/retest/web/selenium/WebElementWrapper.java
```diff
@@ -0,0 +1,131 @@
...
+
+	@Override
+	public void submit() {
+		driver.check( toStep( "submit", delegate ) );
+		delegate.submit();
+	}
+
+	@Override
+	public void sendKeys( final CharSequence... keysToSend ) {
+		driver.check( toStep( "enter", delegate ) );
```
beatngu13: I think it would be helpful if the step would also include the `keysToSend`, what do you think? Potentially too much chars then?

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
beatngu13: `EventFiringWebDriver` also has `EventFiringKeyboard` and `EventFiringMouse` (even `EventFiringTouch`). So I believe we also have to overwrite at least `getKeyboard()` and `getMouse()` and create wrappers for these?

src/test/java/de/retest/web/selenium/AutocheckingRecheckDriverTest.java
```diff
@@ -0,0 +1,19 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+import org.openqa.selenium.remote.RemoteWebDriver;
+
+class AutocheckingRecheckDriverTest {
+
+	@Test
+	void test_makeUnique() {
```
beatngu13: `makeUnique_should_append_unique_prefix`?

src/test/java/de/retest/web/selenium/AutocheckingRecheckDriverTest.java
```diff
@@ -0,0 +1,19 @@
...
+		final AutocheckingRecheckDriver cut = new AutocheckingRecheckDriver( Mockito.mock( RemoteWebDriver.class ) );
```
beatngu13: Static import for `mock`.

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
```diff
@@ -0,0 +1,93 @@
...
+
+	private RecheckWebImpl re;
```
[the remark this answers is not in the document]
roesslerj: No, because we want "unbreakability" baked in.

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
```diff
@@ -0,0 +1,93 @@
...
+
+	public AutocheckingRecheckDriver( final RemoteWebDriver wrapped ) {
```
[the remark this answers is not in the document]
roesslerj: Maybe. I recall there was a reason, but I don't recall which. Anyways, I think this should be another PR then.

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
roesslerj: "stepCounter". Will fix

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
roesslerj: Why is this mandatory? Fits in one line according to formatting rules...

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
roesslerj: Good question. I'd argument that it's not a checking recheck driver that is auto, but an auto-checking recheck driver. See more examples from https://www.grammarly.com/blog/hyphen/

> It’s recommended you don’t take down any load bearing walls when renovating.
>
> In this sentence, it sounds like you shouldn’t take down any load that is holding up a wall. A hyphen should be inserted between load and bearing to make it clear that we’re talking about walls that are bearing a load.
> 
> It’s recommended you don’t take down any load-bearing walls when renovating.
> This rock-hard cake is absolutely impossible to eat.
> We’re looking for a dog-friendly hotel.
>
> You should use the auto-checking driver when you don't want to be bothered with checking at all.

src/main/java/de/retest/web/selenium/WebElementWrapper.java
roesslerj: Thanks, will improve.

src/main/java/de/retest/web/selenium/WebElementWrapper.java
roesslerj: We can shorten to e.g. 10 chars?

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
beatngu13: The (Eclipse) formatter doesn't handle streams very well, which is why we agreed on this: https://retest.atlassian.net/wiki/spaces/RET/pages/201916424/Best+practices

beatngu13:
Regarding 98cd0aa9c76323fa719e95fe7db7271d1784edae: We shouldn't do this dogmatically. I think SonarCloud is the problem here and we need a way to circumvent quality gates when they are not appropriate.

PR #204, opened 2019-04-14
Improve readme logo and add Twitter button and free price tag...

roesslerj:


beatngu13:
I personally think the price badge looks odd, especially on GitHub. I honestly haven't seen this before. You like that?

I would also suggest to align the tweet button with the badges.

Just my two cents … let me know if you want the changes there anyway.

roesslerj:
I found the price badge here: https://github.com/froala/design-blocks. I think it might be helpful, but I am not sure.

How would you align the tweet button?

beatngu13:
>How would you align the tweet button?

Horizontally with the other badges, but we can also keep it there if you want.

roesslerj:
> > How would you align the tweet button?
> Horizontally with the other badges, but we can also keep it there if you want.

I think the tweet button is more of a call to action than the other badges. Therefore I think it makes sense to have it separate.

PR #205, opened 2019-04-15
Feature/autocheckingdriver

roesslerj:


src/main/java/de/retest/web/selenium/ActionbasedCheckNamingStrategy.java
```diff
@@ -0,0 +1,81 @@
...
+			return stringToSend;
+		}
+		return stringToSend.substring( 0, 7 ) + "...";
+	}
+
+	protected String shortenUrl( final Object url ) {
+		if ( url == null ) {
+			return "";
+		}
+		return url.toString().replace( "http://", "" ).replace( "https://", "" );
```
beatngu13: Why would you replace "http" with "https" if the former is actually being used?

src/main/java/de/retest/web/selenium/ActionbasedCheckNamingStrategy.java
```diff
@@ -0,0 +1,81 @@
...
+			// remove trailing ]
+			result = result.substring( 0, result.length() - 1 );
+			// remove identification criterion (e.g. id, class, ...)
+			result = result.substring( result.lastIndexOf( ':' ) + 1 ).trim();
+		}
+		return result;
+	}
+
+	protected String shortenTextInput( final CharSequence[] keysToSend ) {
+		if ( keysToSend == null || keysToSend.length == 0 || keysToSend[0] == null || keysToSend[0].length() == 0 ) {
```
beatngu13: Don't you have to incorporate the entire content of the `keysToSend` array and not just the string on position 0? Same in line 63.

src/main/java/de/retest/web/selenium/AutocheckingCheckNamingStrategy.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Interface to create names for checks for the {@link AutocheckingRecheckDriver}. These names must be unique within a
+ * test, but can repeat in between tests (i.e. after calling {@link #nextTest()}).
+ */
+public interface AutocheckingCheckNamingStrategy {
+
+	/**
+	 * Return a string that is a unique identifier in the current test.
+	 */
+	String getUniqueCheckName( String action, WebElement target, Object... params );
```
beatngu13: To avoid warnings, you have to document all `@param` and `@return`.

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
```diff
@@ -0,0 +1,115 @@
...
+		namingStrategy = new CounterCheckNamingStrategy();
+		options = RecheckOptions.builder().build();
+	}
+
+	// TODO Use RecheckWebOptions
+	public AutocheckingRecheckDriver( final RemoteWebDriver wrapped, final RecheckOptions options,
+			final AutocheckingCheckNamingStrategy namingStrategy ) {
+		super( wrapped );
+		this.options = options;
+		// TODO Incorporate AutocheckingCheckNamingStrategy into RecheckWebBuilder
```
beatngu13: You mean `RecheckWebOptions`?

src/test/java/de/retest/web/selenium/ActionbasedCheckNamingStrategyTest.java
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	@Test
+	void getUniqueCheckName_should_contain_text_for_sendKeys() {
+		final WebElement mock = mock( WebElement.class );
+		when( mock.toString() ).thenReturn(
+				"[[ChromeDriver: chrome on MAC (5822a9b14739d081f70f6b6f42e789cc)] -> link text: Contact]" );
+
+		assertThat( cut.getUniqueCheckName( "enter", mock,
+				(Object[]) new CharSequence[] { "this very long text was sent" } ) )
```
beatngu13: Can't you do:

```java
assertThat( cut.getUniqueCheckName( "enter", mock, "this very long text was sent" ) );
```

?

src/main/java/de/retest/web/selenium/ActionbasedCheckNamingStrategy.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * This class aims to create semantic names for checks based on the previously executed action that led to that state.
+ * This looks like "enter_[my text]_into_user" or "click_contact". If a name occurs multiple times, the second and
+ * following occasions are labeled with "_2", "_3" etc. The drawback of this naming strategy is, that if e.g. the target
+ * changes, the state will not be recognized anymore. So it is recommended to use this naming strategy _only_ in
+ * combination with the <code>retestId</code>.
+ */
+public class ActionbasedCheckNamingStrategy implements AutocheckingCheckNamingStrategy {
+
+	private final Multiset<String> checks = HashMultiset.create();
```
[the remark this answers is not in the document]
roesslerj: True. But I think this is easier and as long as we have Guava in the CP...

src/main/java/de/retest/web/selenium/ActionbasedCheckNamingStrategy.java
roesslerj: I would remove both, as this is a non-info.

src/main/java/de/retest/web/selenium/ActionbasedCheckNamingStrategy.java
roesslerj: Hmm. Not sure about that. In the tests it worked out right.

src/main/java/de/retest/web/selenium/ActionbasedCheckNamingStrategy.java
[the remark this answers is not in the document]
beatngu13: OK, but then we should add it to our dependencies if we actually depend on it.

src/main/java/de/retest/web/selenium/ActionbasedCheckNamingStrategy.java
beatngu13: No, what I mean is that this way we completely ignore sent keys after index 0. I don't know the Selenium API well enough, but if somebody does:

```java
webElement.sendKeys("a", "b", "c");
```

We only use `"a"` but ignore `"b"` and `"c"`?

We could therefore try to use `Arrays.toString( Object[] )`?

src/main/java/de/retest/web/selenium/ActionbasedCheckNamingStrategy.java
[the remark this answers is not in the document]
roesslerj: I disagree. Just for MultiMap, we don't need a specific version. But when explicitly referencing a version, we run into all kinds of problems when it starts to differ from what is referenced transitively. I just had the following problem in Bayern, due to Guava:
``` 10:25:55.014 ERROR [AWT-EventQueue-0] [de.retest.swing.UncaughtExceptionHandlerCrashDetector] - Uncaught exception in thread Thread[AWT-EventQueue-0,6,main]: 
 java.lang.NoSuchMethodError: com.google.common.base.Preconditions.checkArgument(ZLjava/lang/String;I)V
	at com.google.common.hash.AbstractHashFunction.newHasher(SourceFile:75)
	at com.google.common.hash.AbstractHashFunction.hashBytes(SourceFile:65)
	at com.google.common.hash.AbstractHashFunction.hashBytes(SourceFile:59)```

PR #207, opened 2019-04-23
Feature/recheckdriver always all

roesslerj:


src/main/java/de/retest/web/selenium/RecheckDriver.java
```diff
@@ -0,0 +1,29 @@
...
+import org.openqa.selenium.remote.RemoteWebDriver;
+
+import de.retest.recheck.Recheck;
+import de.retest.recheck.RecheckOptions;
+
+/**
+ * This is a wrapper for different {@link RemoteWebDriver} implementations, such as {@link ChromeDriver} or
+ * {@link FirefoxDriver}.
+ *
+ * The RecheckDriver extends AutocheckingDriver, such that it is unnecessary to call
```
beatngu13: Would you mind wrapping both classes in `{@code }` such that they get rendered with a monospaced font?

PR #215, opened 2019-04-26
Feature/i frame handling

beatngu13:
Reopens #212 after fixed rebase/merge conflicts.

@roesslerj please do not approve. Added you as a reviewer so that you get notified by comments.


src/main/java/de/retest/web/selenium/By.java
```diff
@@ -39,6 +40,17 @@ private static Element findElement( final List<Element> children, final Predicat
 		return null;
 	}
 
+	public static List<Element> findElements( final List<Element> children, final Predicate<Element> predicate ) {
```
beatngu13: How about separate unit tests?

src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -61,18 +63,52 @@ public boolean canCheck( final Object toVerify ) {
...
 		}
 
 		return Collections.singleton( lastChecked );
 	}
 
+	public void addChildrenFromFrames( final WebDriver driver, final Set<String> cssAttributes,
+			final RootElement lastChecked ) {
+		final JavascriptExecutor jsExecutor = (JavascriptExecutor) driver;
+		final List<Element> frames = de.retest.web.selenium.By.findElements( lastChecked.getContainedElements(),
+				element -> "iframe".equalsIgnoreCase( element.getIdentifyingAttributes().getType() )
```
beatngu13: If we 

* extract the predicate
* and use an import for `By`

I think it would be a bit more readable:

```java
final Predicate<Element> isFrame = element -> {
    String type = element.getIdentifyingAttributes().getType();
    return Stream.of( "iframe", "frame" ).anyMatch( type::equalsIgnoreCase );
}
final List<Element> frames = By.findElements( lastChecked.getContainedElements(), isFrame );
```

src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -61,18 +63,52 @@ public boolean canCheck( final Object toVerify ) {
...
+						|| "frame".equalsIgnoreCase( element.getIdentifyingAttributes().getType() ) );
+
+		logger.debug( "Found {} frames, getting HTML per frame.", frames.size() );
```
beatngu13: >"[…] getting HTML per frame."

We are not just getting the HTML. Maybe just say "data" or something?

src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -61,18 +63,52 @@ public boolean canCheck( final Object toVerify ) {
...
+		for ( final Element frame : frames ) {
+			final String frameId = frame.getIdentifyingAttributes().get( "id" );
+			if ( frameId == null ) {
+				// TODO Implement handling e.g. via name, XPaht, etc.
+				logger.error( "Cannot retrieve frame with id null from {}.", frame );
```
beatngu13: "ID" in uppercase like below and maybe "data" as well?

```suggestion
				logger.error( "Cannot retrieve data with ID null from frame '{}'.", frame );
```

src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -61,18 +63,52 @@ public boolean canCheck( final Object toVerify ) {
...
+				continue;
+			}
+			try {
+				logger.debug( "Switching to frame with ID {}.", frameId );
+				driver.switchTo().frame( frameId );
+				@SuppressWarnings( "unchecked" )
+				final Map<String, Map<String, Object>> frameAttributesMapping =
+						(Map<String, Map<String, Object>>) jsExecutor.executeScript( getQueryJS(), cssAttributes );
+				final RootElement frameContent = convertToPeers( frameAttributesMapping, "", null );
```
beatngu13: Instead of having no title (empty string), we could reuse the frame ID like:

```suggestion
				final RootElement frameContent = convertToPeers( frameAttributesMapping, "frame-" + frameId, null );
```

src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -61,18 +63,52 @@ public boolean canCheck( final Object toVerify ) {
...
+
+		logger.debug( "Found {} frames, getting HTML per frame.", frames.size() );
+		for ( final Element frame : frames ) {
+			final String frameId = frame.getIdentifyingAttributes().get( "id" );
+			if ( frameId == null ) {
+				// TODO Implement handling e.g. via name, XPaht, etc.
+				logger.error( "Cannot retrieve frame with id null from {}.", frame );
+				continue;
+			}
+			try {
```
beatngu13: Why do we wrap this inside a try-catch, but not the same code in line 67?

src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -61,18 +63,52 @@ public boolean canCheck( final Object toVerify ) {
...
+				logger.debug( "Switching to frame with ID {}.", frameId );
+				driver.switchTo().frame( frameId );
+				@SuppressWarnings( "unchecked" )
+				final Map<String, Map<String, Object>> frameAttributesMapping =
+						(Map<String, Map<String, Object>>) jsExecutor.executeScript( getQueryJS(), cssAttributes );
+				final RootElement frameContent = convertToPeers( frameAttributesMapping, "", null );
+				frame.addChildren( frameContent.getContainedElements() );
+			} catch ( final Exception e ) {
+				logger.error( "Exception retrieving HTML content of frame {}.", frameId, e );
```
beatngu13:
```suggestion
				logger.error( "Exception retrieving data from frame with ID {}.", frameId, e );
```

src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -61,18 +63,52 @@ public boolean canCheck( final Object toVerify ) {
 		logger.info( "Retrieving attributes for each element." );
 		final Set<String> cssAttributes = attributesProvider.getCssAttributes();
 		final JavascriptExecutor jsExecutor = (JavascriptExecutor) driver;
-		@SuppressWarnings( "unchecked" ) final Map<String, Map<String, Object>> result =
+		@SuppressWarnings( "unchecked" )
+		final Map<String, Map<String, Object>> rawAttributesMapping =
```
beatngu13: If below is `frameAttributesMapping`, maybe this should be `windowAttributesMapping`?

src/test/resources/pages/empty-div.html
```diff
@@ -22,5 +22,14 @@
     }
 </script>
 
+<iframe id="myiframe" height="200" frameborder="0" title="description">
```
beatngu13: Page is called `empty-div` and also the test name is `EmptyDivTestIT`. We probably shouldn't mix things up as this now also tests IFrames.

At least rename page and test I would suggest. We can also use different test methods/cases for different aspects.

src/test/java/de/retest/web/selenium/ByTest.java
```diff
@@ -0,0 +1,29 @@
...
+	@Test
+	void findElements_should_return_matching_children() {
+		final Element matching = mock( Element.class );
+		when( matching.toString() ).thenReturn( "matching" );
+		final Element nonMatching = mock( Element.class );
+		final List<Element> children = new ArrayList<>();
+		children.add( matching );
+		children.add( nonMatching );
+
+		Assertions.assertThat( By.findElements( children, child -> child.toString().equals( "matching" ) ) )
```
beatngu13: Should be a static import.

PR #227, opened 2019-05-19
Feature/fix framepath

roesslerj:


src/main/java/de/retest/web/PeerConverter.java
```diff
@@ -81,7 +81,14 @@ private WebElementPeer convertToPeer( final String path, final WebData webData )
 		return peer;
 	}
 
+	protected boolean isRoot( final String parentPath ) {
+		return parentPath == null;
+	}
+
 	static String getParentPath( final String path ) {
+		if ( path.lastIndexOf( '/' ) == -1 ) {
```
beatngu13: -1 means it doesn't occur at all, so I assume we could use:

```suggestion
		if ( !path.contains( "/" ) ) {
```

src/main/java/de/retest/web/mapping/PathsToWebDataMapping.java
```diff
@@ -28,6 +28,12 @@ public PathsToWebDataMapping( final Map<String, Map<String, Object>> mapping ) {
 				.collect( Collectors.toMap( Entry::getKey, entry -> new WebData( entry.getValue() ) ) );
 	}
 
+	public PathsToWebDataMapping( final String path, final Map<String, Map<String, Object>> mapping ) {
```
beatngu13: We could make it more clear what this ctor is made for:

```java
/**
 * @param frameParentPath The parent path of the inline frame.
 * @param mapping The raw map of paths to maps of attributes.
 */
public PathsToWebDataMapping( final String frameParentPath, final Map<String, Map<String, Object>> mapping ) {
    // …
}
```

PR #230, opened 2019-05-22
If the old element cannot be matched to a new one, give better error msg

roesslerj:


src/main/java/de/retest/web/selenium/By.java
```diff
@@ -19,12 +19,16 @@ public static Element findElement( final RootElement lastExpectedState, final Ro
...
 		}
-		final Element result = findElement( lastExpectedState.getContainedElements(), predicate );
-		if ( result == null ) {
+		final Element resultFromExpected = findElement( lastExpectedState.getContainedElements(), predicate );
+		if ( resultFromExpected == null ) {
 			return null;
 		}
 		final Alignment alignment = Alignment.createAlignment( lastExpectedState, lastActualState );
-		return alignment.get( result );
+		final Element resultFromActual = alignment.get( resultFromExpected );
```
beatngu13: Good change, but how about a test for the new behavior?

PR #236, opened 2019-05-24
Fix allowed diff mechanism

oezelenes:


beatngu13:
Can't we also remove lines 27 – 30 as these attributes never land in the actual state? We only use them to compute the rectangles for `outline` and `absolute-outline`. CC @roesslerj

PR #240, opened 2019-05-28
Release v1.1.0

beatngu13:


.retest/recheck.ignore.js
```diff
@@ -12,7 +12,7 @@ function contains(array, key) {
 }
 
 function shouldIgnoreAttributeDifference(element, diff) {
```
IgorAlbert: shouldn't we change the function name to "matches"?

beatngu13: Both works. But yes, we should change this on `master`. Would mind opening a PR?

rebazer:

PR #247, opened 2019-06-03
Use workaround for mockito issue 1606

IgorAlbert:
Mockito-Core version 2.28.2 requires at least ByteBuddy version 1.9.7 to work properly.
A workaround would be to set the ByteBuddy dependency explicitly in our pom. 
I am not sure if there would be any side effects, because Selenium uses ByteBuddy version 1.8.15, which probably caused the problem in the first place.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,9 +156,13 @@
 		<dependency>
 			<groupId>org.mockito</groupId>
 			<artifactId>mockito-core</artifactId>
-			<!-- see https://github.com/mockito/mockito/issues/1606 -->
 			<version>2.23.4</version>
 			<scope>test</scope>
+		</dependency>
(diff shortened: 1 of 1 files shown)


beatngu13:
Besides formatting issues, would you mind pushing directly to the dependabot branch? You can then delete this one here.

beatngu13:
Oh and I assume we transitevly get the ByteBuddy dependency from Selenium? Have you had a look at `mvn dependency:tree`?

IgorAlbert:
Sry for formatting, it looked good locally.
Your assumption is correct. We can exclude Bytebuddy from Selenium so it can use it from Mockito, or we write the Mockito dependency before Selenium.

PR #249, opened 2019-06-04
Use attribute-regex for regular expressions

IgorAlbert:


beatngu13:
Would you mind also updating this in our docs?

PR #252, opened 2019-06-08
Fix retest ID provider usage

beatngu13:
We cannot use a static reference, this causes e.g. in surili that two different retest ID providers are being used—the one being passed to `RecheckSeleniumAdapter` and the globally configured.

Thinking about a hotfix for this … waiting for yet another release slows surili development down. Or we finally start with nightlies.

Oh and better look at it commit-wise, second commit produces ugly diffs due to formatting changes.


roesslerj:
No objections agains a hotfix! :-)

PR #259, opened 2019-06-17
Feature/check disabled ro

roesslerj:


src/main/resources/javascript/getAllElementsByPath.js
```diff
@@ -52,6 +52,19 @@ function addCoordinates(extractedAttributes, node) {
 	}
 }
 
+function getDisabled(node) {
```
beatngu13: Like `isShown` or `isNonEmptyTextNode`:

```suggestion
function isDisabled(node) {
```

PR #261, opened 2019-06-18
Feature/check checked

roesslerj:


src/main/resources/javascript/getAllElementsByPath.js
```diff
@@ -87,6 +87,9 @@ function transform(node) {
 		extractedAttributes[attributeName] = attributeValue;
 	}
 	
+	if (node.checked) {
```
diba1013: Why check? You dont do it for the other attributes...

src/main/resources/javascript/getAllElementsByPath.js
```diff
@@ -87,6 +87,9 @@ function transform(node) {
...
+		extractedAttributes["checked"] = node.checked;
+	}
 	// overwrite empty attributes (e.g. 'disabled')
 	extractedAttributes["disabled"] = isDisabled(node);
 	extractedAttributes["read-only"] = node.readOnly;
```
diba1013: I just noticed... We should take the names from their respective attributes: html - `readonly`

roesslerj: Separate PR ;-)

src/main/resources/javascript/getAllElementsByPath.js
roesslerj: 'checked' is only an attribute of inputs that are of type checkboxes... and that cannot easily be expressed in our defaults.yaml. So we do it that way.

diba1013: Yes, same as readonly, disabled, multiple, ... and other "boolean"-attributes, that are added just via`<tag {$attribute}>`

src/main/resources/javascript/getAllElementsByPath.js
diba1013: Sure, just to put it to your attention.

src/main/resources/javascript/getAllElementsByPath.js
roesslerj: For disabled, readonly and others, we can have default attributes in the defaults.yaml. Because we know for which tags they can appear. But `checked` attribute can appear on some instances of input and not others. So whatever way you put the defaults, there's always an edge case that doesn't fit. So we solve this in the JavaScript. That was actually Daniels idea...

beatngu13: Shouldn't we treat it like the `disabled` attribute? `disabled` is handled via its own method `isDisabled(node)` and will always be `true` or `false`, we just treat `false` as a default so that it's not part of the resulting state. In contrast, checked is `true` or not present at all?

I would suggest to do the same here: Have a method like `isChecked(node)` that always returns `true` or `false` like:

```js
extractedAttributes["checked"] = isChecked(node);
```

Then add `checked: false` to our `defaults.yaml`.

This way, we are not just consistent when it comes to handling possibly absent attributes, the user can configure the behavior via our defaults mechanism.

roesslerj: Hmm. The difference is, that `disabled` is a property of all `input` elements, whereas `checked` is a property only of `input` elements of type `checkbox`. But I do see your point, and we can change that accordingly...

diba1013: I think the idea is not to check for `if(node.checked)`, so that `"checked"` has always a value with either `true` or `false` and cannot contain three states with `undefined`.
If I understand @beatngu13 correctly, he suggest to do something like (via JavaScript-Magic):
```suggestion
	extractedAttributes["checked"] = node.checked || false;
```

beatngu13: As long as it is hidden in separate method (like `isDisabled`, i.e., `isChecked`) and the user can leverage our defaults-mechanism, I'm good. 😉

src/main/resources/javascript/getAllElementsByPath.js
```diff
@@ -65,6 +65,13 @@ function isDisabled(node) {
 	return node.disabled;
 }
 
+function isChecked(node) {
+	if (node.checked) {
+		return true;
+	}
+	return node.false;
```
diba1013: [Well](https://stackoverflow.com/questions/17320961/how-do-i-simplify-an-if-statement-that-returns-true-or-false) .... why not just 
```suggestion
	return node.checked || false;
```
if you are worried about `undefined` see [here](https://stackoverflow.com/questions/21367903/javascript-false-undefined-vs-undefined-false)

PR #263, opened 2019-06-24
Get only mvn args from env variable

modulo11:


ci/deploy.sh
```diff
@@ -8,4 +8,4 @@ set -o pipefail
 export MVN="mvn --settings ${TRAVIS_BUILD_DIR}/.travis.settings.xml"
```
beatngu13: `MVN` can now be removed from all CI scripts.

PR #275, opened 2019-07-19
Feature/use latest firefox

modulo11:


.travis.yml
```diff
@@ -6,7 +6,7 @@ git:
 
 addons:
   chrome: stable
-  firefox: "64.0.2"
+  firefox: latest
```
roesslerj: Isn't this critical to always use the latest firefox? That way, an update of the firefox can break our build. I think I would prefer to keep the version fixed...

modulo11: I disagree because:
- we want to adapt to changes in firefox quickly
- we also use latest chrome
- version 64.0.2 was released on January 9, 2019

roesslerj: Ok.

Issue #295, opened 2019-07-25
Inserted and deleted elements aren't recognized anymore


beatngu13:
Appears to be a regression for which we don't have tests (i.e. verify insertion, change, and deletion yield a test failure).

Steps to reproduce:

1. Execute e.g. `CenterTestIT` and verify it passes locally
2. Delete the `<div>` (ID "center") in lines 35 – 37 from `src/test/resources/pages/centered.html`
3. Run the test again, it passes (although it should fail)

In contrast, if you change the data of the `<div>` (e.g. adapt the text content), then the test fails as expected.

**EDIT:** Inserted elements also seem to be affected. If you simply add another `<div>` to the `<body>`, the test passes too.

roesslerj:
Should be addressed with https://github.com/retest/recheck/pull/343 in recheck 1.3.1.

PR #297, opened 2019-07-25
Support more IFrames

beatngu13:
We are currently limited to frames that have an ID (see https://github.com/retest/recheck-web/blob/master/src/main/java/de/retest/web/RecheckSeleniumAdapter.java#L96). This PR is intended to support all kind of IFrames, i.e., without an ID, a name or something else.

I've marked this as a draft because it is untested, somewhat blocked by #295.

### DIFF
--- a/src/main/java/de/retest/web/FrameConverter.java
+++ b/src/main/java/de/retest/web/FrameConverter.java
@@ -21,11 +21,6 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 public class FrameConverter {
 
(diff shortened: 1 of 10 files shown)


beatngu13:
The currently failing tests are due to the IFrame on our old website, which has no ID (i.e. wasn't properly checked before this): https://travis-ci.com/retest/recheck-web/builds/120787290#L1772

beatngu13:
@roesslerj could you please have a look at this draft?

Issue #290, opened 2019-07-24
Argument is of an illegal type: de.retest.web.selenium.WebElementWrapper


javydreamercsw:
This is related to #288 . Trying to run UnbreakableDriver.executeScript lets to the error below. I assume it's not unwrapping the proxy before passing it down.

Here's the relevant stack trace:
`java.lang.IllegalArgumentException: Argument is of an illegal type: de.retest.web.selenium.WebElementWrapper
	at org.openqa.selenium.remote.internal.WebElementToJsonConverter.apply(WebElementToJsonConverter.java:83)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at org.openqa.selenium.remote.RemoteWebDriver.executeScript(RemoteWebDriver.java:484)
	at de.retest.web.selenium.UnbreakableDriver.executeScript(UnbreakableDriver.java:231)`

I'm using recheck-web 1.3.0.

javydreamercsw:
The call is triggered from a WebDriverEventListener with the following code:

`@Override
  public void beforeClickOn(WebElement arg0, WebDriver arg1)
  {
((JavascriptExecutor) arg1).executeScript("return window.scrollY;", fromElement);
}`

roesslerj:
Thanks for bringing that to our attention.

roesslerj:
Addressed in PR https://github.com/retest/recheck-web/pull/305

roesslerj:
Fixe in 1.3.1.

PR #305, opened 2019-07-31
When we added the method, we also should have added the marker interface

roesslerj:
Should solve issue https://github.com/retest/recheck-web/issues/290

### DIFF
--- a/src/main/java/de/retest/web/selenium/WebElementWrapper.java
+++ b/src/main/java/de/retest/web/selenium/WebElementWrapper.java
@@ -10,8 +10,9 @@ import org.openqa.selenium.Point;
 import org.openqa.selenium.Rectangle;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.WrapsElement;
 
-public class WebElementWrapper implements WebElement {
+public class WebElementWrapper implements WebElement, WrapsElement {
 
 	private final WebElement delegate;
 	private final AutocheckingRecheckDriver driver;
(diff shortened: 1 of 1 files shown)


src/main/java/de/retest/web/selenium/WebElementWrapper.java
```diff
@@ -10,8 +10,9 @@
 import org.openqa.selenium.Rectangle;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.WrapsElement;
 
-public class WebElementWrapper implements WebElement {
+public class WebElementWrapper implements WebElement, WrapsElement {
 
 	private final WebElement delegate;
```
beatngu13: For convenience and consistency, rename to `wrappedElement`?

roesslerj: We can do that, but please in a separate PR. Small steps...

roesslerj: And this is a bugfix, that we can include in a hotfix release. Your suggestion would be a major change that we needed to test again.

beatngu13: Why would that be a "major change"? I'm talking about renaming the field, of which the scope is limited to this class. From:

```java
@Override
public WebElement getWrappedElement() {
	return delegate;
}
```

To:

```java
@Override
public WebElement getWrappedElement() {
	return wrappedElement;
}
```

The getter says `getWrappedElement`, not `getDelegate`. This just follows conventions and avoids e.g. problems with Lombok. IMHO this change belongs in this PR.

roesslerj: Ok, then I misunderstood... Done

PR #316, opened 2019-08-08
Feature/ret 1654 write message about breaking test

ivonx:


diba1013:
I will close this due to inactivity, however keeping the branch so that it may be referenced.

PR #327, opened 2019-08-30
Add an implementation for an AutpcheckingNavigationWrapper

roesslerj:


src/main/java/de/retest/web/selenium/AutocheckingNavigationWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+
+	@Override
+	public void to( final String url ) {
+		delegate.to( url );
+		autocheckingDriver.check( "navigate-to" );
+	}
+
+	@Override
+	public void to( final URL url ) {
+		delegate.to( url );
```
beatngu13: If you want to, you can reuse the method above:

```java
@Override
public void to( final URL url ) {
	to( url.toString() );
}
```

src/main/java/de/retest/web/selenium/AutocheckingCheckNamingStrategy.java
```diff
@@ -21,6 +21,15 @@
...
 	String getUniqueCheckName( String action, WebElement target, Object... params );
 
+	/**
+	 * Return a string that is a unique identifier in the current test.
+	 *
+	 * @param action
+	 *            The action to use when creating a semantic name (e.g. "back").
+	 * @return The unique identifier.
+	 */
+	String getUniqueCheckName( String action );
```
beatngu13: I think you could reuse the other Javadoc:

```java
/**
 * @see #getUniqueCheckName( String, WebElement, Object... )
 */
```

src/main/java/de/retest/web/selenium/CounterCheckNamingStrategy.java
```diff
@@ -16,9 +16,13 @@ public String getUniqueCheckName( final String action, final WebElement target,
 		return String.format( "%02d", stepCounter++ );
```
beatngu13: DRY:

```suggestion
 		return getUniqueCheckName( action );
```

src/main/java/de/retest/web/selenium/AutocheckingNavigationWrapper.java
roesslerj: Hmm. Might make a difference for the delegate, so I'd rather only call that...

src/main/java/de/retest/web/selenium/AutocheckingCheckNamingStrategy.java
roesslerj: Actually I changed the example ("navigate-back" instead of "click").

src/main/java/de/retest/web/selenium/CounterCheckNamingStrategy.java
roesslerj: Done

PR #329, opened 2019-09-02
Improve readme in various ways

roesslerj:
- Add link to Chrome extension
- Change `RecheckDriver` to `UnbreakableDriver` (first one is autochecking)
- Add short explanation how "unkreakable" works

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ In order to use "Unbreakable Selenium", you just need to wrap your usual driver 
  re = new RecheckWebImpl();
 ```
 
-Note that this works only in conjunction with at least one previous call to `re.check`, as behind the scenes, if the element cannot be found on the current page, then recheck-web searches for it in the _last_ Golden Master (where e.g. the ID still is), makes a 1-on-1 assignment to the current elements and returns the element with the highest match, if it's higher than a configurable confidence level. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,21 +154,23 @@ Additionally, a file named `${TEST_CLASS_NAME}.report` will be created upon test
...
 ```java
  // Use the RecheckDriver as a wrapper for your usual driver.
- driver = new RecheckDriver( new ChromeDriver() );
+ driver = new UnbreakableDriver( new ChromeDriver() );
 
  // Use the unbreakable recheck implementation.
  re = new RecheckWebImpl();
 ```
 
+Note that this works only in conjunction with at least one previous call to `re.check`, as behind the scenes, if the element cannot be found on the current page, then recheck-web searches for it in the _last_ Golden Master (where e.g. the ID still is), makes a 1-on-1 assignment to the current elements and returns the element with the highest match, if it's higher than a configurable confidence level.
```
beatngu13:
```suggestion
Note that this works only in conjunction with at least one previous call to `Recheck#check(...)`, as behind the scenes, if the element cannot be found on the current page, then recheck-web searches for it in the _last_ Golden Master (where e.g. the ID still is), makes a 1-on-1 assignment to the current elements and returns the element with the highest match, if it's higher than a configurable confidence level. Also check out the other [ways of using recheck-web](https://docs.retest.de/recheck-web/ways-of-using-recheck-web/).
```

PR #335, opened 2019-09-07
Resolve "Duplicating managed version 2.5.3 for maven-release-plugin"

roesslerj:


pom.xml
```diff
@@ -234,7 +234,6 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-release-plugin</artifactId>
-				<version>2.5.3</version>
```
diba1013: Where do you see this warning? There is no other version specified.

roesslerj: Eclipse. Great IDE, you should try it at some point ;-)

diba1013: Ah, in the effective pom is a section with `pluginManagment` which defines the version&mdash;propably generated by defaults in maven?
I'd prefer to keep the version specified for unison reasons (if the warning does not bug you too much 😄)
@martin-v any preferences?

martin-v: The version is defined in the maven installation, so it's system depended. I would prefer to stay with the version and add a `<!--$NO-MVN-MAN-VER$-->` comment.

PR #338, opened 2019-09-10
Add implementation for individual RemoteWebElements and test

roesslerj:


diba1013:
Did we discuss this to be implemented (now)? @martin-v

beatngu13:
> Did we discuss this to be implemented (now)? @martin-v

I'm surprised, too. I thought we agreed on _not_ doing this—at least for now. Also, @diba1013 already had a prototypical implementation, why a new one? And why not mark this a draft?

roesslerj:
Hmm. Now I am surprised, because we discussed that yesterday during lunch. Sorry you didn't notice that.

martin-v:
@roesslerj the discussion was on the way to lunch only between both of us. And yes I agreed to implement it. But I'm surprised that you have the time to implement and and above all to test it properly ;)
* does it work with review properly?
* does it work if the path changes?
* does it work for inserted and deleted elements? also if the path changes?
* ...

src/main/resources/javascript/getAllElementsByPath.js
```diff
@@ -128,6 +128,27 @@ function containsOtherElements(element) {
 	return element.children.length > 0;
 }
 
+function getElementXPath(node) {
```
diba1013: We really need tests for js&mdash;sooner than later!

src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -49,19 +50,36 @@ public RecheckSeleniumAdapter() {
 
 	@Override
 	public boolean canCheck( final Object toVerify ) {
-		return toVerify instanceof WebDriver;
+		return toVerify instanceof WebDriver || toVerify instanceof RemoteWebElement;
```
diba1013: What about `WrapsElement` and `WrapsDriver` to ensure the widest spectrum possible. 

See: [extract](https://github.com/retest/recheck-web/blob/6d724b7123aeb356d4180cefd35056ab9822512e/src/main/java/de/retest/web/adapter/WebElementRecheckAdapter.java#L52) and [execute](https://github.com/retest/recheck-web/blob/6d724b7123aeb356d4180cefd35056ab9822512e/src/main/java/de/retest/web/mapping/JavaScriptWebDataProvider.java#L115).

src/test/java/de/retest/web/it/RecheckRemoteWebElementIT.java
```diff
@@ -0,0 +1,47 @@
...
+	@MethodSource( "de.retest.web.testutils.WebDriverFactory#drivers" )
+	void simple_html_page_should_be_checked( final WebDriver driver ) throws Exception {
+		this.driver = driver;
+		re.startTest( "simple-page-" + driver.getClass().getSimpleName() );
+
+		driver.get( PageFactory.page( Page.PAGE_FRAME ) );
+		Thread.sleep( 1000 );
+		final WebElement email = driver.findElement( By.id( "email" ) );
+
+		re.check( email, "open" );
```
diba1013: Things that come to mind:
1. What if I insert a html node, does this still work like a driver (without additional window information)
2. What if the node has (no) children
3. What if the node is inserted/deleted
4. Too many changes (e.g. wrong node) since the algorithm has quite the trouble with root changes (i.e html[1] was deleted; html[1] was inserted)
5. All in conjunction with Unbreakable

src/main/java/de/retest/web/PeerConverter.java
```diff
@@ -81,7 +90,7 @@ private WebElementPeer convertToPeer( final String path, final WebData webData )
 	}
 
 	protected boolean isRoot( final String parentPath ) {
-		return parentPath == null;
+		return parentPath == null || parentPath.equals( rootParentPath );
```
diba1013: Do we want this as root, since a [screen and title seems awkward](https://github.com/retest/recheck-web/pull/338/files#diff-2c55711e8071f1b366e12fc1bf319442R4) on a single WebElement. Don't we just want to wrap this into a `RootElement`, as we have a driver.

src/main/resources/javascript/getAllElementsByPath.js
roesslerj: Totally in for that! :-)

src/main/java/de/retest/web/RecheckSeleniumAdapter.java
roesslerj: Very good catch!

src/test/java/de/retest/web/it/RecheckRemoteWebElementIT.java
roesslerj: 1. I think it should. Something surely to test—will add one.
2. There is a test for no children. Will add one with children.
3. User responsibility (is `null`?)
4. We'll have to see about that in practice.
5. Well this simply doesn't work with the UnbreakableDriver. To have test be unbreakable, you need to give the complete context.

src/main/java/de/retest/web/PeerConverter.java
roesslerj: I agree. But I think this is a problem to be addressed in the data model (i.e. in the `RootElement`) rather than here. We should not make another bad decision because of an earlier bad decision...

src/main/java/de/retest/web/mapping/PathsToWebDataMapping.java
```diff
@@ -39,18 +37,14 @@ public PathsToWebDataMapping( final Map<String, Map<String, Object>> mapping ) {
 	 *            The raw map of paths to maps of attributes.
 	 */
 	public PathsToWebDataMapping( final String frameParentPath, final Map<String, Map<String, Object>> mapping ) {
-		mapping.forEach( ( k, v ) -> findRoot( k ) );
+		rootPath = frameParentPath + mapping.keySet().stream().reduce(
+				( rootPath, path ) -> countMatches( path, "/" ) < countMatches( rootPath, "/" ) ? path : rootPath )
+				.orElse( "" ).replace( "//", "/" );
```
diba1013: I preferred the sorting stuff..., since you essentially do Bubble Sort.

src/main/java/de/retest/web/RootElementPeer.java
```diff
@@ -31,11 +31,20 @@ public RootElement toElement( final Element parent ) {
 			throw new IllegalStateException( "RootElement was not properly initialized!" );
 		}
 		final IdentifyingAttributes identifyingAttributes = retrieveIdentifyingAttributes();
+
+		// If this is a WebElement
+		String screen = title;
+		String title = this.title;
+		if ( !identifyingAttributes.getType().equals( "html" ) ) {
```
diba1013: Well, if it is a single element and a html (see your test)? What do I expect.

src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -49,19 +51,40 @@ public RecheckSeleniumAdapter() {
 
 	@Override
 	public boolean canCheck( final Object toVerify ) {
-		return toVerify instanceof WebDriver;
+		return toVerify instanceof WebDriver || toVerify instanceof RemoteWebElement //
```
beatngu13: Just a short note from my side: We currently mix `WebDriver` and `RemoteWebElement` in the public API. Why not have both `WebDriver` and `WebElement` or `RemoteWebDriver` and `RemoteWebElement`?

diba1013: Well, the clean solution would be to have a adapter each (i. e. driver and element separated). The only part of the RemoteElement is the `WrapsDriver` interface (of which the driver is hopefully a `JavaScriptExecutor`). So technically we would need a check for `instanceof WebElement & WrapsDriver`. 

My implementation does this and provides a fallback to gather the information directly from the element.

However, as we want(/need) this out as fast as possible for the release, I settled with this (not so clean) implementation.

beatngu13: >as we want(/need) this out as fast as possible for the release

It is public API, so I am pro investing some more time to come up with a proper solution. Once it's merged, it's gonna stay and we have to maintain it. `WebDriver` and `RemoteWebElement` just doesn't feel right.

In addition, many people use the `WebDriver` inteface on the left-hand side as their type, just have a look at the official Selenium API introduction:

https://www.seleniumhq.org/docs/03_webdriver.jsp#introducing-the-selenium-webdriver-api-by-example

This will only return `WebElement`s, so that wouldn't work with this API without casting. In contrast, `RemoteWebElement` implements `WebElement`, so this signature would be compatible.

diba1013: I agree that we should loosen the `RemoteWebElement` to a `WebElement` and check similar to: [extract](https://github.com/retest/recheck-web/blob/6d724b7123aeb356d4180cefd35056ab9822512e/src/main/java/de/retest/web/adapter/WebElementRecheckAdapter.java#L52) and [execute](https://github.com/retest/recheck-web/blob/6d724b7123aeb356d4180cefd35056ab9822512e/src/main/java/de/retest/web/mapping/JavaScriptWebDataProvider.java#L115).

That means, we only work with `WebElement`, `WrapsElement` and `WrapsDriver`.

PR #341, opened 2019-09-12
Add Changelog starting with 1.4.0

roesslerj:


CHANGELOG.md
```diff
@@ -0,0 +1,53 @@
...
+
+--------------------------------------------------------------------------------
+
+
+[1.5.0] (2019-09-12)
+--------------------
+
+### Improvements
+
+* Check now also individual `WebElement`s, making recheck a "Jest for Java".
```
beatngu13: Hmmm, snapshot testing is just a small part of Jest. It also has a test runner, mocking capabilities, coverage tracker, … moreover, difference testing is different than snapshot testing as we don't diff text. You really wanna keep it? I find it a bit confusing.

beatngu13:
Did you forget to add a reviewer?

CHANGELOG.md
roesslerj: If someone enters "Jest for Java" as a search term, I want recheck to show up. There are differences, but still... I think in the Changelog, it doesn't hurt...

beatngu13: Hmmm, I don't think that the changelog is the right place to do SEO … here we want to give clear information to our users what has changed, no marketing blabber.

Just my two cents, feel free to merge anyway.

PR #342, opened 2019-09-13
Feature/recheck cap

Creepow:
Reports get uploaded after every single test because cap() is called after each test. Maybe change the lifecyle of recheck so cap() gets only called after every test was executed. That requires a recheck instance to create only once.
There is also the issue with AutocheckingRecheckDriver to execute re.cap() in the quit() method, which may need to be changed as well.
[See this ticket](https://retest.atlassian.net/browse/RET-1798)

### DIFF
--- a/src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
+++ b/src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
@@ -77,8 +77,7 @@ public class AutocheckingRecheckDriver extends UnbreakableDriver {
 	@Override
 	public void close() {
 		// Is this sensible? What about tests using separate sessions?
-		namingStrategy.nextTest();
-		re.cap();
+		cap();
 		super.close();
 	}
 
@@ -86,8 +85,7 @@ public class AutocheckingRecheckDriver extends UnbreakableDriver {
 	public void quit() {
 		try {
 			// Is this sensible? What about tests using separate sessions?
-			namingStrategy.nextTest();
-			re.cap();
+			cap();
 		} finally {
 			super.quit();
 		}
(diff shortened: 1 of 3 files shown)


beatngu13:
While this is actually the correct lifecycle, I suggest to wait for the JUnit extension(s) by @briemla. Otherwise, the tests are getting a bit verbose.

We also have to discuss how we handle the issue with our driver implementations. Here, `WebDriver` lifecycle != `Recheck` lifecycle.

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
```diff
@@ -77,17 +77,15 @@ public void get( final String url ) {
 	@Override
 	public void close() {
 		// Is this sensible? What about tests using separate sessions?
-		namingStrategy.nextTest();
-		re.cap();
```
roesslerj: Hmm. I don't understand that change. Maybe ask Daniel for review, or explain to me on Monday...

beatngu13: We reuse the local `cap()` method ([see here](https://github.com/retest/recheck-web/blob/a690bb0b63d635e9964921fe864305d0294ab482/src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java#L66)), which does the same.

rebazer:

rebazer:

rebazer:

beatngu13:
Close until merge conflicts have been resolved. @Creepow could you do this please?

Creepow:
Merge conflicts resolved. Looks clean.

PR #343, opened 2019-09-13
Bump recheck

beatngu13:


beatngu13:
Fails because not available on central (yet).

beatngu13:
I have found the issue, which comes from `TestReportFilter`:

https://github.com/retest/recheck/blob/release/v1.5.0/src/main/java/de/retest/recheck/report/TestReportFilter.java#L112

`Element#hasAnyDifference()` doesn't include the child differences, so every difference gets eventually filtered with this approach.

I see two options:

1. Include child differences in `Element#hasAnyDifference()` and preserve the hierarchy of differences, although parents may be empty. Then, the corresponding printer(s) have to filter these empty parents.
2. Let `TestReportFilter` remove all empty parents and only preserve actual differences, breaking the hierarchy of differences.

CC @diba1013

PR #344, opened 2019-09-16
Feature/simple page diff test

Creepow:


src/test/java/de/retest/web/testutils/PageFactory.java
```diff
@@ -31,7 +31,7 @@ public static String page( final Page page ) {
 		}
 	}
 
-	private static final String toPageUrlString( final String relativePath ) {
+	public static final String toPageUrlString( final String relativePath ) {
```
beatngu13: The corresponding commit message "Change PageFactory to public" doesn't really explain why you did this. From our best practices (via https://chris.beams.io/posts/git-commit/):

>7. Use the body to explain what and why vs. how

In the body you should write some notes on _why_ you did this, e.g., "To allow reuse in other tests."

Issue #326, opened 2019-08-23
Provide a way to select elements to monitor


javydreamercsw:
I'm loving the tool but sometimes feels like is trying to do too much leading to too much noise.

It would be awesome if we could annotate WebElements within a PageObject to be monitored (meaning screenshots taken as it's being interacted with) ignoring the to non-monitored ones.

At work there are multiple regions working on different parts of the complex application that we don't care about in some focused tests.

Looks like this goes against #29. In our case the current behavior generates way too many files already.

roesslerj:
Hey! Thanks for the feedback. We're glad you like it. 😃

We're not sure we understood what your challenge is. Could we maybe have a short session via Skype, hangout, Jitsi, ... were you show us your problem? Then it'd be much easier for us to conceive a solution strategy.

Thanks!

diba1013:
If I understand you correctly, you want to focus only on specific `WebElements` (i.e. `<div></div>`, etc.) while ignoring the noise inside the GUI.

### Current way

An alternative, that is **currently** available, is to define a custom filter or ignore-rules (which we currently have no documentation for!?), but essentially boils down to [this](https://docs.retest.de/recheck/how-ignore-works/).

A ignore can be defined in `${project-root}/.retest/recheck.ignore` with the [rules](https://docs.retest.de/recheck/how-ignore-works/) mentioned above. There you can specify globally for all tests, which elements you do not want to monitor.

If you want to have this only for a single test (advanced), we are currently working on #182 which is partly implemented for [suites ](https://github.com/retest/recheck/pull/193) while we wait for the support for [steps](https://github.com/retest/recheck-web/issues/182). See [this comment](https://github.com/retest/recheck-web/issues/182#issuecomment-488715207) on how it should work.

A filter can be defined in `${project-root}/.retest/filter/${name-of-filter}.filter` which is displayed in **_review_** or can be referenced inside **_recheck.cli_** while diffing. It is a way to reduce the noise within the GUI and temporarily removing all differences for which the filter applies. Thus here you can define&mdash;similarly to the ignore-mechanism&mdash;which regions to ignore.

_We are happy to help and to extend our documentation as needed._

### Proposal

[...]

PR #348, opened 2019-09-16
Fix/deprecated ctor

roesslerj:
Otherwise we're not compatible, as pointed out here https://github.com/retest/recheck-web/issues/326#issuecomment-531230689


diba1013:
Fixes #349

PR #371, opened 2019-09-25
Rethrow exception on getWrapped

diba1013:
since it causes Stackoverflow through loop if the same element is returned every time.

This is an issue when working with PageObjects (i.e. `@FindBy`) which do the evaluation of the actual element lazy (via [`Proxy`](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Proxy.html)) in `getWrapped`. Thus the construction of the PageObject is fine, until you access the object and its `WrapsElement` and a `NoSuchElementException` or similar is thrown.

However, the current implementation just silences this exception and instead returns the same element, which results in a recursion with no exit, and thus ultimately a `StackOverFlowError`.

This PR addresses this issue by rethrowing any potential exception from `getWrapped` with a `RuntimeException` and breaking the loop if the method could not be called resulting in an `UnsupportedOperationException`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,12 +27,22 @@ Table of Contents
 
 ### Bug Fixes
 
+* Fix `StackOverflowError` if a `WrapsElement` (e.g. `@FindBy`) is checked that throws an exception.
+
 ### New Features
 
 ### Improvements
 
 
 --------------------------------------------------------------------------------
+
+
+[1.5.1] (2019-09-23)
+--------------------
+
+### Bug Fixes
+
+* Fix [#355](https://github.com/retest/recheck-web/issues/355) when an exception is thrown in JavaScript if the element has no parent. 
 
 
 [1.5.0] (2019-09-12)
--- a/src/main/java/de/retest/web/util/SeleniumWrapperUtil.java
+++ b/src/main/java/de/retest/web/util/SeleniumWrapperUtil.java
@@ -44,12 +44,20 @@ public class SeleniumWrapperUtil {
 	}
 
 	/**
+	 * Extracts the wrapped element or driver from a given object while implementing backwards compatibility to the
+	 * internal Selenium API. It is advised to call {@link #isWrapper(WrapperOf, Object)} before calling this method.
+	 * 
 	 * @param w
 	 *            The wrapper type.
 	 * @param o
 	 *            The object to be used.
-	 * @return The wrapped element or driver from the given object if it is instance of the selected type, otherwise the
-	 *         object itself.
+	 * @return The wrapped element or driver from the given object if it is instance of the selected type.
+	 * 
+	 * @throws RuntimeException
+	 *             If the <code>getWrapped</code> method throws an exception.
+	 * @throws UnsupportedOperationException
+	 *             If the <code>getWrapped</code> method cannot be invoked or the object does not wrap an instance of
+	 *             the selected type.
 	 */
 	public static Object getWrapped( final WrapperOf w, final Object o ) {
 		final Class<?> clazz = getWrapperClass( w, o );
--- a/src/main/java/de/retest/web/util/SeleniumWrapperUtil.java
+++ b/src/main/java/de/retest/web/util/SeleniumWrapperUtil.java
@@ -1,5 +1,8 @@
 package de.retest.web.util;
 
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+
 import lombok.Getter;
 
 /**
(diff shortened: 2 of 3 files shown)


src/main/java/de/retest/web/util/SeleniumWrapperUtil.java
```diff
@@ -41,21 +44,37 @@ public static boolean isWrapper( final WrapperOf w, final Object o ) {
...
+				throw new RuntimeException(
+						"Failed to invoke " + o.getClass().getSimpleName() + "#" + w.getWrapperMethodName() + ".",
+						e.getTargetException() );
+			} catch ( final NoSuchMethodException | IllegalAccessException e ) {
+				throw new UnsupportedOperationException(
+						"Failed to invoke " + o.getClass().getSimpleName() + "#" + w.getWrapperMethodName() + ".", e );
+			}
 		}
-		return o;
+		throw new UnsupportedOperationException( "Type '" + o.getClass() + "' is not instance of any of "
```
beatngu13: Rather than having this at the end—which is a very good find by the way—how about an early return:

```java
final Class<?> clazz = getWrapperClass( w, o );
if ( clazz == null ) {
    throw new IllegalArgumentException( ... ) // I think this fits better.
}
try {
    // ...
}
```

src/test/java/de/retest/web/util/SeleniumWrapperUtilTest.java
```diff
@@ -66,4 +66,28 @@ void should_get_wrapped_driver() throws Exception {
 		assertThat( SeleniumWrapperUtil.getWrapped( WrapperOf.DRIVER, oldWrapper ) ).isSameAs( wrapped );
 	}
 
+	@Test
+	void getWrapped_should_not_silently_ignore_exceptions_thrown_by_method() {
+		final WrapsElement elementThrows = mock( WrapsElement.class );
+		when( elementThrows.getWrappedElement() ).thenThrow( new NoSuchElementException( "No element found" ) );
+
+		assertThatCode( () -> SeleniumWrapperUtil.getWrapped( WrapperOf.ELEMENT, elementThrows ) )
```
beatngu13: Do you know what is the difference to `assertThatThrownBy`?

src/main/java/de/retest/web/util/SeleniumWrapperUtil.java
diba1013: Yes, `IllegalArgumentException` is definitely better.

src/test/java/de/retest/web/util/SeleniumWrapperUtilTest.java
diba1013: Firstly, I would have said nothing, since they have the same return. But looking at the code `assertThatThrownBy` verifies that an exception has been thrown. That means that `assertThatCode` can expect no assertion thrown.

Since I expect a exception, I will change it

beatngu13: Didn't know this too, thx for looking it up. 👍

diba1013:
> By the way: We now have a changelog! 😉

Almost forgot ...

CHANGELOG.md
```diff
@@ -35,6 +37,14 @@ Table of Contents
 --------------------------------------------------------------------------------
 
 
+[1.5.1] (2019-09-23)
```
beatngu13: Ooops … thanks! 🙈

CHANGELOG.md
```diff
@@ -35,6 +37,14 @@ Table of Contents
...
+--------------------
+
+### Bug Fixes
+
+* Fix #355 when an exception is thrown in JavaScript if the element has no parent.
```
beatngu13: Unfortunately, GitHub doesn't seem to automatically create hyperlinks for this.

```suggestion
* Fix [#355](https://github.com/retest/recheck-web/issues/355) when an exception is thrown in JavaScript if the element has no parent. 
```

src/main/java/de/retest/web/util/SeleniumWrapperUtil.java
```diff
@@ -41,21 +44,37 @@ public static boolean isWrapper( final WrapperOf w, final Object o ) {
 	}
 
 	/**
+	 * Extracts the wrapped element or driver from a given object while implementing backwards compatibility to the
+	 * internal selenium api. It is advised to call {@link #isWrapper(WrapperOf, Object)} before calling this method.
```
beatngu13: Minor/optional (sorry, forgot previously):

```suggestion
	 * internal Selenium API. It is advised to call {@link #isWrapper(WrapperOf, Object)} before calling this method.
```

diba1013:
@beatngu13 Could you disable this exception rule, so that rebazer can merge this?

PR #374, opened 2019-09-26
Scroll down to bottom

roesslerj:
Since we create a snapshot of the whole website, we should scroll down
first, to trigger all sorts of loading mechanisms...

### DIFF
--- a/src/main/java/de/retest/web/RecheckSeleniumAdapter.java
+++ b/src/main/java/de/retest/web/RecheckSeleniumAdapter.java
@@ -102,13 +102,14 @@ public class RecheckSeleniumAdapter implements RecheckAdapter {
 	}
 
 	private Set<RootElement> convert( final WebDriver driver, final RemoteWebElement webElement ) {
+		// Do not inline this, as we want the screenshot created before retrieving elements
+		final BufferedImage screenshot = shoot( driver, webElement );
 		final Set<String> cssAttributes = attributesProvider.getCssAttributes();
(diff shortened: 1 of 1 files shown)


beatngu13:
Scrolling up and down caused issues with Surili in the past. And what happens when scrolling is possibly endless (e.g. Twitter feed)?

We also already do this in `RecheckSeleniumAdapter` when creating the screenshot with Shutterbug—but it is [configurable via a property](https://docs.retest.de/recheck-web/properties/). However, this is appears to be not necessary for all browser.

Nevertheless, we shouldn't a) do this twice (JS and Shutterbug) and b) without giving the user the power to toggle this feature. Therefore, I suggest to not merge the PR in its current state and think about a more flexible solution.

roesslerj:
Very good comments, thank you. Actually, what we should do instead then, is to create the screenshot first, and _then_ retrieve the DOM.

I'll change accordingly.

src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -102,13 +102,13 @@ public boolean canCheck( final Object toVerify ) {
 	}
 
 	private Set<RootElement> convert( final WebDriver driver, final RemoteWebElement webElement ) {
+		final BufferedImage screenshot = shoot( driver, webElement );
```
beatngu13: Good idea, this should work. 👍 But maybe we should add a comment that the order matters? Or adapt the `shoot` method name that it possibly scrolls? Otherwise, one may change the order or inline the variable again?

PR #375, opened 2019-09-26
Feature/root element is null

roesslerj:


roesslerj:
Commit looks good to me. Have you tried undoing the fix to make sure the test detects this? If yes, you can approve :-)

diba1013:
> Commit looks good to me. Have you tried undoing the fix to make sure the test detects this? If yes, you can approve :-)

Of course 😄

PR #378, opened 2019-10-02
Update readme

beatngu13:
Just some minor fixes as follow-up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,24 +221,25 @@ If you cannot easily the reports on your CI/CD server, test reports can be easil
 
 The first step is to modify the `setUp()` method in our existing test case to enable the upload to ***rehub***. There are two ways to achieve this:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -221,24 +221,27 @@ If you cannot easily the reports on your CI/CD server, test reports can be easil
 
 The first step is to modify the `setUp()` method in our existing test case to enable the upload to ***rehub***. There are two ways to achieve this:
```
roesslerj: Just realized ... this doesn't necessarily have to be in the `setUp` method... do we want to change that?

beatngu13: You mean the system property? I think it's fine as an example, we already provide alternative approaches here: https://docs.retest.de/recheck-web/tutorial/upload-test-reports-to-rehub/#enable-rehub-globally

PR #381, opened 2019-10-14
Add an show case test for the RecheckDriver and mention it in the docs

roesslerj:


README.md
```diff
@@ -110,7 +110,7 @@ A complete tutorial about how to setup and use ***recheck-web*** like this can e
 
 ### Implicit Checks
 
-You can also use the recheck driver wrapper, which will automatically create a check after every action. To use this, simply wrap your driver like so:
+You can also use the recheck driver wrapper, which will automatically create a check after every action. To use this, [simply wrap your driver like so](https://github.com/retest/recheck-web/blob/master/src/test/java/de/retest/web/it/RecheckDriverShowcaseIT.java):
```
beatngu13: Why not `AutocheckingRecheckDriver`, as this section is about implicit checks? We already have a test for it too:

https://github.com/retest/recheck-web/blob/master/src/test/java/de/retest/web/it/SimpleAutocheckingDriverShowcaseIT.java

Then we can also avoid having another expensive test that slows down the build.

roesslerj: Because a) we want to reference it directly in the readme (see change) and b) we want to also make sure, that the RecheckDriver still works.
The test is only expensive, because that branch: https://github.com/retest/recheck-web/tree/feature/omitScreenshots isn't in yet. Screenshot creation is SOOO expensive. We need to get rid of it in our existing tests...

roesslerj: And that branch depends on https://github.com/retest/recheck/tree/feature/move-props-to-opts...

beatngu13: Regarding the readme: the section is about implicit checks. Implicits checks are what we have `AutocheckingRecheckDriver` for. `RecheckDriver` simply aggregates all features. Therefore, I would argue that in this section we should link against `SimpleAutocheckingDriverShowcaseIT`.

Regarding screenshots: while screenshot creation might be extra expensive, Selenium tests are also slow without it. We talk about seconds. I also think that in its current form, `RecheckDriver` isn't worth this price because it is essentially the same as `AutocheckingRecheckDriver`. But if you really, really want to, you can do it.

Regarding the other branch: how does this relate to this PR?

roesslerj: I think we should not link potential users to the AutocheckingDriver, but to the RecheckDriver. That should be what users of the API use. Thus we should have a test for it and link to it.
The other branch only relates to your comment about how expensive the tests are. Btw. I did measure before cutting. Screenshot creation is about 100x as expensive as the rest...

beatngu13: >I think we should not link potential users to the AutocheckingDriver, but to the RecheckDriver. That should be what users of the API use.

Frankly, that wasn't obvious to me. If this is this case, I'm fine with it. But then I guess we should change the Javadoc [here](https://github.com/retest/recheck-web/blob/master/src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java) and point to the usage of `RecheckDriver` and the docs [here](https://docs.retest.de/recheck-web/ways-of-using-recheck-web/) too.

>Btw. I did measure before cutting. Screenshot creation is about 10x-100x as expensive as the rest...

Again, that wasn't point: even if you have an ordinary Selenium tests with recheck, screenshots, etc. they are expensive—this is also why people try to avoid them (take the test pyramide for example).

So, I'm good with the new test, but shall we keep `SimpleAutocheckingDriverShowcaseIT`? It does essentially same, and since it is not intended to be used directly, we at least need no showcase anymore?

beatngu13: @roesslerj?

roesslerj: How would you change the Javadoc?

roesslerj: And I think since we want to support both, we need tests for both.

beatngu13: >How would you change the Javadoc?

Have a visible note that says something like:

```
Please don't use this class directly, instead use {@link RecheckDriver}.
```

>And I think since we want to support both, we need tests for both.

One test takes 10s on my shiny new Mac, why do the very same thing twice? `RecheckDriver` equals `AutocheckingDriver` in terms of behavior. If we have a test for the former, we automatically test the latter. Therefore, I have argued that I would suggest to drop `SimpleAutocheckingDriverShowcaseIT` for `SimpleRecheckDriverShowcaseIT`.

If you really, really want to, keep it. I just don't think it's worth slowing down the build.

roesslerj: > Please don't use this class directly, instead use {@link RecheckDriver}.
These classes serve different purposes. I wouldn't want to add that, it is misleading. And yeah, we slow down the build, but this just emphasizes that we need to address the performance issue. We need to feel the pain, too.

beatngu13: >These classes serve different purposes. I wouldn't want to add that, it is misleading.

But you said:

>I think we should not link potential users to the AutocheckingDriver, but to the RecheckDriver. That should be what users of the API use.

Then, what is the purpose of these classes? What is part of the public API? What should users use? What should developers use?

At this point, I'm super confused and can't tell what to use when. As soon as this is clear, I think we can (finally) merge this thing.

roesslerj: Hmm. Then maybe it's me. Let's discuss this in person on Wednesday...

diba1013:
Until this discussion has been resolved, I will close this. 

I agree that we should strive for the `RecheckDriver` to be used. But as of now there is no difference between `AutocheckingDriver` and `RecheckDriver` and thus no real ground to discuss this on. 

To properly resolve this, I think we should firstly discuss [these points](https://github.com/retest/recheck-web/pull/470#issue-355166222), especially the need for the `AutocheckingDriver` being unbreakable. I would argue that this does not need to be the case and is now possible that we can distinguish between internal and external usage. That would give the `AutocheckingDriver` its identity and thus reason to be used, (i.e. I do not need the major overhead of unbreakable but just want the autochecking). If it should be unbreakable (as it is currently), I would completely remove either of the classes (`AutocheckingDriver` vs `RecheckDriver`, in favor of the last due to not speaking naming) in the next major release.

Issue #166, opened 2019-02-25
Improve RecheckSeleniumAdapter#convert( Object ) performance


beatngu13:
Conversion sometimes takes up to 5 or more seconds. It would be helpful (for surili as well) if we could somehow improve the conversion performance.

beatngu13:
This simply test (which uses `convert` under the hood) shows the problem:

```java
class GoogleIT {

	WebDriver driver;
	Recheck re;

	@BeforeEach
	void setUp() throws Exception {
		driver = new ChromeDriver();
		re = new RecheckImpl();
	}

	@Test
	void test() throws Exception {
		re.startTest();
		driver.get( "https://policies.google.com/privacy?hl=de&gl=de" );
		final long start = System.currentTimeMillis();
		re.check( driver, "german-privacy-policy" );
		final long end = System.currentTimeMillis();
		System.out.println( "check took " + TimeUnit.MILLISECONDS.toSeconds( end - start ) + " seconds" );
		re.capTest();
	}

	@AfterEach
	void tearDown() throws Exception {
		driver.quit();
		re.cap();
	}

}
```

This single check takes up to 14 seconds on my machine.

beatngu13:
12 seconds for https://en.wikipedia.org/wiki/President_of_the_United_States.

diba1013:
We need to do some more investigation on where the actual performance bottleneck is.

- No Golden Master, vs Golden Master present.
- Retrieve data from Selenium (i.e. JavaScript)
- Convert Performance
- Diffing Algorithm 
- Persisting Golden Master

PR #397, opened 2019-11-01
Use the new RecheckOptions mechanism to omit screenshot creation

roesslerj:
Screenshot creations takes up the vast majority of the time

### DIFF
--- a/src/main/java/de/retest/web/RecheckWebOptions.java
+++ b/src/main/java/de/retest/web/RecheckWebOptions.java
@@ -1,5 +1,10 @@
 package de.retest.web;
 
+import java.awt.image.BufferedImage;
+
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+
 import de.retest.recheck.RecheckOptions;
 import de.retest.web.selenium.AutocheckingCheckNamingStrategy;
 import de.retest.web.selenium.CounterCheckNamingStrategy;
@@ -17,10 +22,23 @@ public class RecheckWebOptions extends RecheckOptions {
 	 */
 	private final AutocheckingCheckNamingStrategy checkNamingStrategy;
 
+	/**
+	 * The {@link AttributesProvider} that defines the CSS attributes to check.
+	 */
+	private final AttributesProvider attributesProvider;
+
+	/**
+	 * The {@link ScreenshotProvider} to use.
+	 */
+	private final ScreenshotProvider screenshotProvider;
+
 	public RecheckWebOptions( final RecheckOptions superOptions,
-			final AutocheckingCheckNamingStrategy checkNamingStrategy ) {
+			final AutocheckingCheckNamingStrategy checkNamingStrategy, final AttributesProvider attributesProvider,
+			final ScreenshotProvider screenshotProvider ) {
 		super( superOptions );
 		this.checkNamingStrategy = checkNamingStrategy;
+		this.attributesProvider = attributesProvider;
+		this.screenshotProvider = screenshotProvider;
 	}
 
 	public static RecheckWebOptionsBuilder builder() {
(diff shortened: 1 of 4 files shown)


src/main/java/de/retest/web/RecheckWebOptions.java
```diff
@@ -30,16 +48,33 @@ public static RecheckWebOptionsBuilder builder() {
...
 			this.checkNamingStrategy = checkNamingStrategy;
 			return this;
 		}
 
+		public RecheckWebOptionsBuilder attributesProvider( final AttributesProvider attributesProvider ) {
+			this.attributesProvider = attributesProvider;
+			return this;
+		}
+
+		public RecheckWebOptionsBuilder omitScreenshots() {
```
diba1013: I think the word `omit` is mostly used in a negative context (see [here](https://www.lexico.com/en/synonym/omit))

> Leave out or exclude (someone or something), either intentionally or forgetfully.

I would propose `noScreenshots` or `disableScreenshots`.

roesslerj:
> As this does not have any priority, I would merge this into the master and not the ongoing release.

Hmm. People are really concerned about performance. recheck-web _is_ slow—however, 95% of the time gets lost in screenshot creation. I feel that this is an issue we need to address as fast as possible, i.e. in this release (this branch was ready 2 weeks ago, see our other discussion).

I think how we handle the ScreenshotProvider should be an internal decision, i.e. people should be able to configure it with an abstract option, like here. So I think this doesn't depend on #399. 

retest/recheck#432 is a very good point and something that we missed—but also an internal discussion and solution.

So my proposal would be to have this in 1.6.0 (will fix changelogs, tests, documentation), then address the other issues. Or turned on its head: if we have this in the current release, does this prevent us from addressing #399 or retest/recheck#432 as proposed?

diba1013:
> this branch was ready 2 weeks ago, see our other discussion

Luckily, GitHub has a feature for that: [Draft PRs](https://github.blog/2019-02-14-introducing-draft-pull-requests/). We would then have had the time to discuss necessary changes and address them accordingly. Of course this PR firstly would have had to be merged into the master first, changing the base to release once finished.

I want to do the release as fast as possible (ideally finishing it today or tomorrow) and not wait for a new feature that might as well be in the 1.7.0 once we ironed out the issues.

>  If we have this in the current release, does this prevent us from addressing #399 or retest/recheck#432 as proposed

Well, yes, since we expose the `ScreenshotProvider` as public API via `RecheckOptions#getScreenshotProvider`. Thus changes to the `ScreenshotProvider` would introduce breaking changes, at least with #399. 

Secondly, extending a class and changing its behavior&mdash;as done with `RecheckWebOptionsBuilder#omitScreenshots`&mdash;goes against the contract of extending (see [here](https://codeburst.io/inheritance-is-evil-stop-using-it-6c4f1caf5117)). Interfaces are for providing services, classes for implementing such service.

> Hmm. People are really concerned about performance. recheck-web is slow—however, 95% of the time gets lost in screenshot creation.

I'd love to see the study behind these tests (e.g. documented within #166 or our wiki). I know, that we have serious performance issues within recheck(-web). However, Selenium is not fast as well and removing an important feature, just to improve performance seems not the first way to go. Screenshots are essential and used to identify the differences within a screen and an major selling point for review. You see the difficulties when trying to accept differences within recheck.cli where you have no clue on what elements are affected by a change.

We recently changed the screenshots in #374 to be an essential part of stabilizing the page. Did you check (and wrote tests) that this does not alter the behavior (i.e. introduce unexpected differences just because the screenshot is gone)? How does it compare for driver and element?

---

I get that the user can still decide if the screenshots are necessary. However, for the above mentioned reasons, I would at least propose (as done before) to split this PR into two, so that we have time to discuss how we want to address the screenshots.

diba1013:
We will postpone this into the next release.

PR #399, opened 2019-11-01
Always resize the image to max-width of currently 800px

roesslerj:
This is already implemented in the Chrome extension, where we needed to reduce the image size as this was _the_ major performance bottleneck of the backing micro service. The general idea is simply that we don't need the image other than showing what changed to the user. Therefore we can increase performance and decrease the needed space and bandwidth simply by having the screenshot downscaled. Any objections?


diba1013:
This goes in conjunction with #397 and RET-1843. However, we should adapt the API of `ScreenshotProvider` and expose this inside the options, such that this can be configured.

```
interface ScreenshotProvider {
    BufferedImage shootBrowser(WebDriver) throws Exception;
    BufferedImage shootElement(WebDriver, Element) throws Exception;
}
```

Then we could have a 
* `NopScreenshotProvider`
* `FullScreenshotProvider`
* `ViewportOnlyScreenshotProvider`
* `CappedScreenshotProvider` with the maxWidth, which acts upon another provider
* ...

This could be done iteratively.

Secondly, we should not mix the viewport (or screen) and the window. The window is the actual browser, the screen is the representation of the browser (i.e. the screenshot). The model should act on the window coordinates and not have any clue about the screen coordinates.. Thus the elements should reflect the actual window size (which we might have to save somewhere). When we mark the screenshots, we then can transform the window coordinates to the screen coordinates and draw them into the screnshot (since this transformation can be easily done).

Ideally any screenshot manipulation should be done with the GPU if available. However, I do not know how this can be achieved within Java.

src/main/java/de/retest/web/ScreenshotProvider.java
```diff
@@ -14,8 +13,12 @@
 import com.assertthat.selenium_shutterbug.core.Shutterbug;
 import com.assertthat.selenium_shutterbug.utils.web.ScrollStrategy;
 
+import de.retest.recheck.ui.image.ImageUtils;
+
 public class ScreenshotProvider {
 
+	/** Should be in sync with the same variable in the JS, in order to have the markings match the elements... **/
```
beatngu13: Either use Javadoc or a simple single-line comment.

src/main/java/de/retest/web/ScreenshotProvider.java
```diff
@@ -14,8 +13,12 @@
...
+	private static final int WANTED_WIDTH = 800;
```
beatngu13: Unit? Maybe `WANTED_WIDTH_PX`?

src/main/java/de/retest/web/ScreenshotProvider.java
```diff
@@ -29,10 +32,10 @@ private ScreenshotProvider() {}
 	public static BufferedImage shoot( final WebDriver driver, final WebElement element ) {
 		try {
 			if ( element != null ) {
-				return shootElement( driver, element );
+				return resizeImage( shootElement( driver, element ) );
 			}
 			final boolean viewportOnly = Boolean.getBoolean( VIEWPORT_ONLY_SCREENSHOT_PROPERTY );
-			return viewportOnly ? shootViewportOnly( driver ) : shootFullPage( driver );
+			return viewportOnly ? resizeImage( shootViewportOnly( driver ) ) : resizeImage( shootFullPage( driver ) );
```
beatngu13:
```java
final BufferedImage img = viewportOnly ? shootViewportOnly( driver ) : shootFullPage( driver );
return resizeImage( img );
```

src/main/java/de/retest/web/ScreenshotProvider.java
```diff
@@ -47,11 +50,15 @@ private static BufferedImage shootFullPage( final WebDriver driver ) {
 		final BufferedImage image = Shutterbug
 				.shootPage( driver, ScrollStrategy.WHOLE_PAGE, SCROLL_TIMEOUT_MS, USE_DEVICE_PIXEL_RATIO ).getImage();
 		return driver instanceof ChromeDriver
-				? resizeImage( image, image.getWidth() / SCALE, image.getHeight() / SCALE ) : image;
+				? ImageUtils.resizeImage( image, image.getWidth() / SCALE, image.getHeight() / SCALE ) : image;
```
beatngu13: Find it a bit confusing that you directly call `ImageUtils` here, but otherwise via the private method.

src/main/java/de/retest/web/ScreenshotProvider.java
```diff
@@ -47,11 +50,15 @@ private static BufferedImage shootFullPage( final WebDriver driver ) {
...
 	}
 
 	private static BufferedImage shootViewportOnly( final WebDriver driver ) {
 		return Shutterbug.shootPage( driver, USE_DEVICE_PIXEL_RATIO ).getImage();
 	}
 
+	private static BufferedImage resizeImage( final BufferedImage image ) {
+		final double height = image.getHeight() * (WANTED_WIDTH / image.getWidth());
+		return ImageUtils.resizeImage( image, WANTED_WIDTH, (int) height );
```
beatngu13: Have measured how expensive this is? Recently we've been quite critical when it comes to screenshots and performance. While resizing reduces the file size, i.e., less local storage needed, faster upload to rehub, … but how does this affect the actual performance of recheck-web?

src/main/java/de/retest/web/ScreenshotProvider.java
roesslerj: Done

src/main/java/de/retest/web/ScreenshotProvider.java
roesslerj: Done

src/main/java/de/retest/web/ScreenshotProvider.java
roesslerj: Different now anyways...

src/main/java/de/retest/web/ScreenshotProvider.java
roesslerj: Done differently anyways...

src/main/java/de/retest/web/ScreenshotProvider.java
roesslerj: It's a one-time operation with many-time benefits. If it's too slow, user should disable screenshots.

src/main/java/de/retest/web/screenshot/ViewportOnlyScreenshot.java
```diff
@@ -1,21 +0,0 @@
-package de.retest.web.screenshot;
```
diba1013: Since this class is public API, we need to preserve this.

src/main/java/de/retest/web/screenshot/ViewportOnlyMinimalScreenshot.java
```diff
@@ -0,0 +1,35 @@
...
+
+import de.retest.recheck.ui.image.ImageUtils;
+
+/**
+ * The default implementation that takes a screenshot of only the viewport visible to the user from the top page.
+ */
+public class ViewportOnlyMinimalScreenshot implements ScreenshotProvider {
+
+	static final int DEFAULT_WANTED_WIDTH_PX = 800;
+	private static final String RESIZE_MAX_WIDTH_PX = "de.retest.recheck.web.resizeMaxWidthPx";
```
diba1013: I assume, we keep this internal for now?
```suggestion
	private static final String RESIZE_MAX_WIDTH_PX = "de.retest.recheck.web.screenshot.maxWidthPx";
```

roesslerj: Yeah, would do that.

diba1013:
I think this looks good. Any objections @beatngu13?

PR #401, opened 2019-11-02
=== seems to strong of a comparator

roesslerj:
Fix: TypeError: Cannot read property "expected" from undefined in <eval> at line number 14


roesslerj:
```
	if (diff.expected == null) {
		print("Expected: " + diff.expected);
		diff.expected = "";
		print("Expected: " + diff.expected);
	}
```
result:
```
Expected: null
Expected: null
```

roesslerj:
Will have to work on this a bit more...

PR #422, opened 2019-11-11
Improve message to be actionable for users

roesslerj:
This is a very simple PR, that only changes the text of an exception. I deem no test should be adapted for this, and I don't think it warrants a changelog entry or documentation change.

### DIFF
--- a/src/main/java/de/retest/web/selenium/NoElementWithHighEnoughMatchFoundException.java
+++ b/src/main/java/de/retest/web/selenium/NoElementWithHighEnoughMatchFoundException.java
@@ -1,5 +1,6 @@
 package de.retest.web.selenium;
 
+import de.retest.recheck.Properties;
 import de.retest.recheck.ui.descriptors.Element;
 
 public class NoElementWithHighEnoughMatchFoundException extends RuntimeException {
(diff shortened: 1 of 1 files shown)


src/main/java/de/retest/web/selenium/NoElementWithHighEnoughMatchFoundException.java
```diff
@@ -8,6 +9,9 @@
 
 	public NoElementWithHighEnoughMatchFoundException( final Element resultFromExpected ) {
 		super( "The element " + resultFromExpected
-				+ " from the Golden Master could not be matched with high enough confidence." );
+				+ " from the Golden Master could not be matched with high confidence > "
```
diba1013: Is it `>` or `>=` 😄

src/main/java/de/retest/web/selenium/NoElementWithHighEnoughMatchFoundException.java
```diff
@@ -8,6 +9,9 @@
...
+				+ System.getProperty( Properties.ELEMENT_MATCH_THRESHOLD_PROPERTY )
+				+ ". To change required confidence, please set property "
+				+ Properties.ELEMENT_MATCH_THRESHOLD_PROPERTY );
```
diba1013: If we "expose" this property now, we should add it to the [docs](https://docs.retest.de/recheck/usage/configuration/#properties) and the `defaults.properties` (although we do not read this currently).

diba1013: Since this is a exception for the user and not for us, should the exception contain following points?

1. Cause: "The element changed too much"
2. Conclusion: "The element 'input' from the Golden Master could not be matched with high enough confidence." 		
3. Solution: "Please adjust the property"

src/main/java/de/retest/web/selenium/NoElementWithHighEnoughMatchFoundException.java
roesslerj: It is `>`. See Alignment:150.

src/main/java/de/retest/web/selenium/NoElementWithHighEnoughMatchFoundException.java
roesslerj: Hmm. If we were certain that it's because the element changed too much, we could just go on and match it. The thing is ... it might be a different element altogether. So I think the current message is good. I'll improve wording a bit, but I won't change it drastically...

roesslerj: Wouldn't want to put it into defaults.properties. But I documented it in https://github.com/retest/docs/pull/60...

PR #423, opened 2019-11-11
Update GM for `example.report`

beatngu13:


beatngu13:
@diba1013 it's still there, but it looks like it has changed?

diba1013:
Looks like a keycloak issue? since [my tests](https://travis-ci.com/diba1013/docs-test) also fail.

PR #426, opened 2019-11-16
If we didn't address the performance, at least we need to tell people …

roesslerj:
…we know how to fix it...

Please approve whoever sees it first :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,11 @@ The [naming strategy](https://github.com/retest/recheck/blob/master/src/main/jav
 You can open the `report` using [***review***](https://retest.de/review/) or the open-source [***recheck.cli***](https://github.com/retest/recheck.cli/). Both let you easily accept changes with a single click (patent pending) or a single command. This also saves a lot of time during maintenance. Moreover, any regularly changing aspect or element (e.g. date fields) can easily be ignored (see [above](#help-i-have-too-many-changes-what-do-i-do)).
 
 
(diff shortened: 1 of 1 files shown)


diba1013:
We will do a new release Soon&trade;. Part of #424 will be to document this, thus I do not think this is necessary, as this PR would have to be reverted partly.

roesslerj:
We want to be fast. This _MUST_ be one of our key goals.
Therefore we must change things fast. This is one of the reasons we have rebazer and automation pipeline.
It _DO NOT ACCEPT_ that we have to discuss a mere 53 word, non-risk, documentation change. If you found a bug (which would be a spelling error), please comment. Otherwise please _APPROVE_ as we need to move forward. If #424 addresses and documents this, great. But until then (and you said it yourself "soon"), we need to get _SHIT DONE_ and at the _very least_ tell people what is going on.

roesslerj:
This behavior in general makes PRs burdensome, which is _the opposite_ of what we want. We want fast, smooth, small changes, as little overhead as possible and we want people to take action and fix stuff. We need to talk about this culture of stall immediately.

beatngu13:
>It [sic!] DO NOT ACCEPT that we have to discuss a mere 53 word, non-risk, documentation change.

Calm down. This is not how we want to deal with reviewer comments:

https://github.com/google/eng-practices/blob/master/review/developer/handling-comments.md

As you said, it is a documentation change. We can figure out a solution.

PR #442, opened 2019-11-27
Move the CSS attributes into the JS file for easier handling

roesslerj:

Only internal change, so no:
- you updated the changelog.
- you updated necessary documentation within [retest/docs](https://github.com/retest/docs).


roesslerj:
Well ... this is what I needed to unify the JS. For me, we can also remove the AttributesProvider. Shall I?

beatngu13:
I cannot think of a use case for it anymore, everything shall be controlled by the JS I guess?

roesslerj:
You added it for Surili. I don't think it was every used outside that. So I'll remove it.

PR #443, opened 2019-11-28
Feature/recheck web properties

Creepow:

- [ ] the code follows the [Clean Code](https://clean-code-developer.com/) guidelines.
- [ ] your commit history is cleaned up.

## Description
Draft PR to decide whether or not we use owner to manage our properties.
> TL;DR 

Needs the change in recheck.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,3 @@
 
 # Artifact from signing JARs.
 retest-gmbh-gpg.asc
-
-# Artifacts from recheck.
-.retest/retest.properties
--- a/.retest/recheck.ignore
+++ b/.retest/recheck.ignore
@@ -24,9 +24,6 @@ matcher: type=meta, attribute-regex: .*
 # Have the tests pass locally:
 attribute=outline
 
-# This should be default:
-attribute=absolute-outline
-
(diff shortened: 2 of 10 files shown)


beatngu13:
Adapted this PR based on https://github.com/retest/recheck/pull/538.

The only thing we still have to think about: There is now `RecheckProperties` and RecheckWebProperties`. While recheck is working with the former, recheck-web uses latter. But in this case, do their live two instances?

diba1013:
> While recheck is working with the former, recheck-web uses latter. But in this case, do their live two instances

Instead of treating them as extension (i.e. `RecheckWebProperties extends RecheckProperties`) and since extension does not really work with owner, we could treat them as real separate properties that just happen to be read from the same file.

beatngu13:
The updated recheck snapshot now includes Rhino, which fails here:

https://travis-ci.com/retest/recheck-web/builds/139369773#L3426

beatngu13:
@diba1013 @roesslerj please note https://github.com/retest/recheck-web/pull/443/commits/77a013fe5434bef62e22f068b5ab850ed0d42517.

PR #451, opened 2019-12-11
Refactor property to how it is documented

Creepow:

- [x] you updated necessary documentation within [retest/docs](https://github.com/retest/docs).

## Description

> TL;DR

### DIFF
--- a/src/main/java/de/retest/web/RecheckWebProperties.java
+++ b/src/main/java/de/retest/web/RecheckWebProperties.java
@@ -35,7 +35,7 @@ public interface RecheckWebProperties extends Reloadable {
 	 * Properties, their key constants and related functionality.
 	 */
 
-	static final String SCREENSHOT_PROVIDER_PROPERTY_KEY = "de.retest.recheck.web.screenshotProvider";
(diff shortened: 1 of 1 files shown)


beatngu13:
Ah, I see! Well, if we documented it as:

```
de.retest.recheck.web.screenshot.provider
```

But internally it was:

```
de.retest.recheck.web.screenshotProvider
```

It didn't work for our users, or they didn't use it (since nobody complained). So, while we are at it: do we also want to fix the package name?

```
de.retest.web.screenshot.provider
```

At least this is what has been implemented and also the automatic module name. @roesslerj @diba1013?

diba1013:
I would have preferred, if all our recheck-web stuff was in the `de.retest.recheck.web` package. Maybe that's why I originally documented this property like it is now. But if we want to stick to the package name instead of what I think the functional name, we can go and do that.

Our properties do not follow any of these conventions, as seen in:
```
de.retest.recheck.project.root=null
de.retest.recheck.elementMatchThreshold=0.3
de.retest.recheck.web.screenshot.provider=viewPortOnly
```

So I do not care that much.

PR #470, opened 2019-12-19
Feature/prevent implicit explicit mixing

diba1013:

- [x] you updated necessary documentation within [retest/docs](https://github.com/retest/docs). See #retest/docs#75.

## Description

> TL;DR Disallow mixing of implicit and explicit checking.

This PR aims to forbid explicit and implicit checking as described here:

```java
@ExtendWith( RecheckExtension.class )
class LoginTest {

	Recheck re;
	WebDriver driver;

	@BeforeEach
	void setUp() {
		re = new RecheckImpl();
		driver = new RecheckDriver( new ChromeDriver() ); 
	}

	@AfterEach
	void tearDown() {
		driver.quit();
	}

	@Test
	void mixing_implicit_and_explicit_checking() throws Exception {
		driver.get( "https://example.com" );

		re.check( driver, "mixing" ); // Will now throw an exception
	}
}
```

#### How does this work?

The `AutocheckingDriver` simply wraps itself (i.e. the UnbreakableDriver) into a separate wrapper `ImplicitDriverWrapper`. Within the `SeleniumAdapter` the wrapper is unwrapped and the original driver is used to be converted. However, the user does not have access to said wrapper and can only pass the `AutocheckingDriver` into the check method. This can now be detected, as the `SeleniumAdapter` simply disallows any raw `AutocheckingDriver` to be passed into the `convert` method as it expects it to be wrapped in the `ImplicitDriverWrapper`.

However that brought some complications which have been resolved, but rise the below mentioned questions. The `RecheckWebImpl` (both explicitly and implicitly used) needs to access the `UnbreakableDriver` behind the `AutocheckingDriver` (which now is wrapped). Therefore it, too, needs to unwrap the driver and pass the wrapped along to the adapter.

### State of this PR

While this PR prevents the mixing, we should discuss some points, which may be addressed separately.

1. Does the `AutocheckingDriver` have to be unbreakable? If I recall correctly from our last discussion about the implicit/explicit topic, this was a limitation in order for the feature to work correctly. 

   However, since we can now wrap any `WebDriver` within `ImplicitDriverWrapper` this should be no limitation anymore. 

    1. The `AutocheckingDriver` can simply call `ImplicitDriverWrapper.of( internalDriver )` and not be unbreakable.
    2. The `RecheckDriver` can simply delegate using `AutocheckingDriver` and pass the `UnbreakableDriver`.

   Secondly, this would resolve the discussion around #381 as each driver would have its own purpose (and the `RecheckDriver` combining all of them).
2. Do we really want to completely block of implicit and explicit mixing. A use case would be to have an unstable page. Consider the following example for a login page where an invalid password has been entered:

    1. Type user `"admin"`.
    2. Type password `"invalid"`.
    3. Click login.
    4. Wait for page to be stable, because the login needs to query the database first.
    5. Expect an error message to be displayed.

    The hindering part is the waiting for stabilization. The login check would still display the spinning button, waiting for a response of the server. How is the user supposed to check for the message being displayed without executing another action? 

   Some solutions come to mind:

   1. Use explicit checking. ... Oh wait, this does not work. However, we could expose the internal driver (e.g.: `re.check( autoRe.getWrappedDriver(), "implicit-and-explicit-working"))`). However, this would only be possible if the above has been addressed.
   2. Have support to wait for stabilization. Selenium does not provide a good way to do so (even though there is a `WebDriverWait`, but requires the driver and a `List<ExpectedConditions>`).

### DIFF
--- a/src/main/java/de/retest/web/RecheckSeleniumAdapter.java
+++ b/src/main/java/de/retest/web/RecheckSeleniumAdapter.java
@@ -85,22 +85,38 @@ public class RecheckSeleniumAdapter implements RecheckAdapter {
 
 	@Override
 	public Set<RootElement> convert( final Object toVerify ) {
-		if ( SeleniumWrapperUtil.isWrapper( WrapperOf.ELEMENT, toVerify ) ) {
-			return convert( SeleniumWrapperUtil.getWrapped( WrapperOf.ELEMENT, toVerify ) );
+		final RemoteWebElement webElement = retrieveWebElement( toVerify );
+		if ( webElement != null ) {
+			return convertWebElement( webElement );
 		}
-		if ( toVerify instanceof RemoteWebElement ) {
-			return convertWebElement( (RemoteWebElement) toVerify );
-		}
-		if ( toVerify instanceof UnbreakableDriver ) {
-			return convertWebDriver( (UnbreakableDriver) toVerify );
-		}
-		if ( SeleniumWrapperUtil.isWrapper( WrapperOf.DRIVER, toVerify ) ) {
-			return convert( SeleniumWrapperUtil.getWrapped( WrapperOf.DRIVER, toVerify ) );
-		}
-		if ( toVerify instanceof RemoteWebDriver ) {
-			return convertWebDriver( (RemoteWebDriver) toVerify );
+		final WebDriver webDriver = retrieveWebDriver( toVerify );
+		if ( webDriver != null ) {
+			return convertWebDriver( webDriver );
 		}
 		throw new IllegalArgumentException( "Cannot convert objects of type '" + toVerify.getClass().getName() + "'." );
+	}
+
+	private RemoteWebElement retrieveWebElement( final Object toVerify ) {
+		if ( SeleniumWrapperUtil.isWrapper( WrapperOf.ELEMENT, toVerify ) ) {
+			return retrieveWebElement( SeleniumWrapperUtil.getWrapped( WrapperOf.ELEMENT, toVerify ) );
+		}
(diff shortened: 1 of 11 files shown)


src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -85,22 +87,50 @@ public boolean canCheck( final Object toVerify ) {
...
-			return convertWebElement( (RemoteWebElement) toVerify );
+			return (RemoteWebElement) toVerify;
+		}
+		return null;
+	}
+
+	private Object unwrapImplicitDriver( final Object toVerify ) {
+		if ( toVerify instanceof AutocheckingRecheckDriver ) {
+			throw new UnsupportedOperationException( String.format(
+					"Mixing implicit checking used by '%s' and explicit checking with 'Recheck#check' is not supported.",
```
roesslerj: This error message is not "actionable". Can we improve the error message, i.e. tell the user how to proceed from here? Examples could be a link to the docs or FAQs or similar.

roesslerj: Alternative error message: "The AutocheckingRecheckDriver does automatically check after each action, therefore no explicit call to re.recheck() is needed. For more information, please have a look at <link to docs>."

diba1013: Why didn't I think of this? This is not a message for us but for the user, therefore it should be actionable. I addressed this with a link to the documentation. Please be aware that the above mentioned docs changes must be merged in order for the link to get valid.

PR #485, opened 2020-01-20
Changed wording in logging from "Found 0 frame(s)" to "Found 0 sub-frame(s)"

roesslerj:
As people might think we didn't get any data if there are no frames. Was a real question/concern...

### DIFF
--- a/src/main/java/de/retest/web/FrameConverter.java
+++ b/src/main/java/de/retest/web/FrameConverter.java
@@ -36,7 +36,7 @@ public class FrameConverter {
 	public void addChildrenFromFrames( final WebDriver driver, final RootElement lastChecked ) {
 		final List<Element> frames =
 				de.retest.web.selenium.By.findElements( lastChecked.getContainedElements(), isFrame() );
-		log.debug( "Found {} frame(s), getting data per frame.", frames.size() );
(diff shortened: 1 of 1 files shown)


src/main/java/de/retest/web/FrameConverter.java
```diff
@@ -36,7 +36,7 @@ public FrameConverter( final String queryJs, final RetestIdProvider retestIdProv
 	public void addChildrenFromFrames( final WebDriver driver, final RootElement lastChecked ) {
 		final List<Element> frames =
 				de.retest.web.selenium.By.findElements( lastChecked.getContainedElements(), isFrame() );
-		log.debug( "Found {} frame(s), getting data per frame.", frames.size() );
+		log.debug( "Found {} sub-frame(s), getting data per frame.", frames.size() );
```
beatngu13: Omit the confusing log entirely if there are no frames?

```java
if ( frames.size() > 0 ) {
    log.debug( "Found {} frame(s), getting data per frame.", frames.size() );
}
```

roesslerj: Hmm. Dann aber if für die ganze Methode, sonst wird das unverständlich.

PR #519, opened 2020-02-06
Feature/Remove unnecessary dependency

lemoncurry:
- [ ] <s>the necessary tests are either created or updated.</s>
- [ ] <s>you updated the changelog.</s>
- [ ] <s>you updated necessary documentation within [retest/docs](https://github.com/retest/docs).</s>

## Description

> Remove unused dependency 

- remove unnecessary dependency org.seleniumhq.selenium:3.141.59
- use direct dependencies instead

### Additional Context

[Maven dependency management recommendations](http://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html)
RET-1964


pom.xml
```diff
@@ -108,7 +108,19 @@
 
 		<dependency>
 			<groupId>org.seleniumhq.selenium</groupId>
-			<artifactId>selenium-java</artifactId>
+			<artifactId>selenium-api</artifactId>
+			<version>3.141.59</version>
```
githubert: If the selenium versions always have to be the same for all three dependencies, you could use something like
```suggestion
			<version>${selenium.version}</version>
```
and then define a property `<selenium.version>3.141.59</selenium.version>` in the `properties` section.

pom.xml
```diff
@@ -130,6 +142,25 @@
 			<version>${jackson.version}</version>
 		</dependency>
 
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-lang3</artifactId>
+			<version>3.9</version>
+			<scope>compile</scope>
```
githubert: You can drop this, as it is the default.
```suggestion
```

lemoncurry:
> State of this PR says WIP, is this correct?

No, sry I was waiting for the checks to pass.

PR #525, opened 2020-02-14
Fix screenshots

### LINKED ISSUE #523: RecheckWebProperties getInstance() called before init()
## Describe the Bug

`RecheckWebProperties#init()` must be called before `RecheckWebProperties#getInstance()` to ensure that the source `"file:${projectroot}/.retest/retest.properties"` can be resolved. We currently do this in `RecheckSeleniumAdapter#initialize( RecheckOptions )`.

However, the `RecheckWebOptionsBuilder` accesses it before:

```java
@Override
public RecheckWebOptions build() {
	return new RecheckWebOptions( super.build(), checkNamingStrategy, screenshotProvider != null
			? screenshotProvider : RecheckWebProperties.getInstance().screenshotProvider() );
}
```

As a consequence, the `retest.properties` aren't loaded yet. This is also why our own reports for this project contain screenshots, although we set the screenshot provider to `none`.

### How to Reproduce?

```java
// for some test in recheck-web
RecheckWebOptions opts = RecheckWebOptions.builder().build();
System.out.println( opts.getScreenshotProvider() );
// de.retest.web.screenshot.ViewportOnlyMinimalScreenshot@XXXXXXXX
```

Although this is normal usage (create options before a `Recheck` instance) , it will ignore the `retest.properties` and just use the default.

### Setup

- recheck-web: 1.9.0
- Selenium version: 3.141.59
- Testing framework: JUnit5 5.6.0
- OS: macOS 10.15.2
- Java version: OpenJDK 8

beatngu13:

## Description

Fixes #523.

### State of this PR

Any idea for a good test case? Some integration between `RecheckWebOptions` and `RecheckSeleniumAdapter`.

Also, build fails due to #524 and changelog needs to be updated.

### DIFF
--- a/src/main/java/de/retest/web/RecheckSeleniumAdapter.java
+++ b/src/main/java/de/retest/web/RecheckSeleniumAdapter.java
@@ -117,8 +117,8 @@ public class RecheckSeleniumAdapter implements RecheckAdapter {
 	private Object unwrapImplicitDriver( final Object toVerify ) {
 		if ( toVerify instanceof AutocheckingRecheckDriver ) {
 			throw new UnsupportedOperationException( String.format(
-					"The '%s' does implicit checking after each action, " // 
-							+ "therefore no explicit check with 'Recheck#check' is needed. " // 
+					"The '%s' does implicit checking after each action, " //
+							+ "therefore no explicit check with 'Recheck#check' is needed. " //
 							+ "Please remove the explicit check. " //
 							+ "For more information, please have a look at https://docs.retest.de/recheck-web/introduction/usage/.",
 					toVerify.getClass().getSimpleName() ) );
--- a/src/main/java/de/retest/web/RecheckSeleniumAdapter.java
+++ b/src/main/java/de/retest/web/RecheckSeleniumAdapter.java
@@ -72,7 +72,6 @@ public class RecheckSeleniumAdapter implements RecheckAdapter {
 
 	@Override
 	public RecheckAdapter initialize( final RecheckOptions opts ) {
-		RecheckWebProperties.init();
 		return new RecheckSeleniumAdapter( opts );
 	}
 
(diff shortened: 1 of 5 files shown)


src/main/java/de/retest/web/RecheckSeleniumAdapter.java
```diff
@@ -53,12 +53,17 @@
...
+	}
+
+	private ScreenshotProvider getScreenshotProvider( final RecheckOptions options ) {
 		if ( options instanceof RecheckWebOptions ) {
-			final RecheckWebOptions webOptions = (RecheckWebOptions) options;
-			screenshotProvider = webOptions.getScreenshotProvider();
-		} else {
-			screenshotProvider = RecheckWebProperties.getInstance().screenshotProvider();
+			final ScreenshotProvider fromOptions = ((RecheckWebOptions) options).getScreenshotProvider();
+			if ( fromOptions != null ) {
```
diba1013: I quite liked the fact that the `RecheckOptionsBuilder` does the check so that the `RecheckOptions` are properly initialized. That was the whole point around the discussion on who should do the initialization: Builder or Options.

This introduces the same problem as with `reportUploadEnabled` which needs the second evaluation for the properties to determine the actual value. Thus it would be the same with the `RecheckOptions`, if moving the report upload to the builder. 

Thus I would suggest to move the init methods within the options, as they are the first to be called. Or maybe into the `RecheckWebProperties#getInstance`?

As a side note. `RecheckOptions#init` and `RecheckWebOptions#init` does the same. As `ConfigFactory` is a static class, wouldn't one of them suffice?

beatngu13: Will resolve this as discussed in Slack.

beatngu13:
Still can't think of a good integration test … how could we verify that the correct (in our case `none` instead of the default `viewportOnly`) screenshot provider is being used? Check the logs?

beatngu13:
Yep, see:

https://github.com/retest/recheck-web/pull/525#issuecomment-586374521

Any idea?

PR #528, opened 2020-02-24
Fix test for compatibility with mockito 3.3.0

githubert:

- [ ] ~the code follows the [Clean Code](https://clean-code-developer.com/) guidelines.~
- [ ] ~you updated the changelog.~
- [ ] ~you updated necessary documentation within [retest/docs](https://github.com/retest/docs).~

## Description

Mockito 3.3.0 will complain about unused stubbings, which in this case seems like a bug. Converting this test into a parameterized test will work around this problem.

### State of this PR

All done.

### Additional Context

https://github.com/retest/recheck-web/pull/527 is currently broken due to a change in mockito.

### DIFF
--- a/src/test/java/de/retest/web/selenium/css/HasTest.java
+++ b/src/test/java/de/retest/web/selenium/css/HasTest.java
@@ -74,8 +74,15 @@ class HasTest {
 		);
 	}
 
-	@Test
-	void should_match_attribute_with_value_containing_word() throws Exception {
+	@MethodSource( "attributeMatchersContainingWord" )
+	@ParameterizedTest( name = "[{index}] using selector ''{0}'', to match ''{1}'', expecting {2}" )
+	void should_match_attribute_with_value_containing_word( final String selector, final String value,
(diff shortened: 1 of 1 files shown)


src/test/java/de/retest/web/selenium/css/HasTest.java
```diff
@@ -74,8 +74,15 @@ void should_match_attribute_with_value_containing_substring( final String select
 		);
 	}
 
-	@Test
-	void should_match_attribute_with_value_containing_word() throws Exception {
+	@MethodSource( "attributeMatchersContainingWord" )
+	@ParameterizedTest( name = "[{index}] using selector ''{0}'', to match ''{1}'', expecting {2}" )
+	void should_match_attribute_with_value_containing_word( final String selector, final String matchingValue,
```
briemla: `matchingValue` no longer describes only the matching value. I would recommend to rename it.

githubert: I changes it to simply say `value`.

PR #530, opened 2020-02-26
Feature/fix unbreakable throws exception for element

diba1013:

- [x] ~you updated necessary documentation within [retest/docs](https://github.com/retest/docs).~

## Description

> TL;DR Elements cannot be passed into RecheckWebImpl as discussed [here](https://github.com/retest/recheck/issues/661#issuecomment-591451731)

This fixes the issue of not being able to pass elements into `RecheckWebImpl`. As it tries to identify the `UnbreakableDriver` from the called object, which obviously fails to identify the driver from an element. Thus an exception is thrown for `loadExpected`.

To fix this, the `RecheckWebImpl` will now query `WrapsDriver` to try to identify the driver within other wrapped drivers. In theory this should work for `RemoteWebElements` as returned by any driver, since they implement `WrapsDriver`. However, the driver is not the `UnbreakableDriver` but the `ChromeDriver`.

This can be resolved, by returning a custom web element that implements both `WrapsElement` and `WrapsDriver`. The former to allow the `SeleniumRecheckAdapter` to find the wrapped element (as returned by the `ChromeDriver`) and the a latter allows the `RecheckWebImpl` to find the driver (`UnbreakableDriver`).

### State of this PR

I did not implement that for the `AutocheckingDriver` or the `RecheckDriver` as both should never pass an element (although in theory, this should work since the extend from `UnbreakableDriver`) and the user should never manually call `re.check(autocheckingDriver.findElementBy(By.id("foo"))` (although we do not explicitly forbid this for elements as of yet...).

In theory, this should not break any existing tests. However, there seem currently tests breaking of which I assume that they are from a browser update. I will need to investigate that before approval.

### Additional Context

This was discovered by retest/recheck#661.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Table of Contents
 * Fix that cascading frames inside of frames are not analyzed. Now depth doesn't matter...
 * Fix that `NoScreenshot` or `RecheckWebOptions#disableScreenshots()` is not applied to web elements.
 * Fix that `de.retest.recheck.web.screenshot.provider` from `recheck.ignore` is ignored when `RecheckWebOptions` are used.
+* Fix that passing an element into `RecheckWebImpl` from an `UnbreakableDriver` throws an error.
 
 ### New Features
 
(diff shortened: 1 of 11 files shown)


diba1013:
I have no idea why there are differences, but they seem to be some minor fluctuations similar to a browser update. Since we for now disabled the report upload until we resolved some issues with the API key, I will need to find a way around it, but keep this PR open so I do not need to recreate it.

diba1013:
I was able to find the fix for the failing tests. While initially, I assumed a browser update I found out that this is actually an inserted action due to wrapping the `AutocheckingWebElement` twice. I was able to fix that and it should be working now.

@roesslerj The PR is now ready for review.

PR #532, opened 2020-02-28
Fix/update recheck

diba1013:


src/test/java/de/retest/web/it/RecheckRemoteWebElementFailingIT.java
```diff
@@ -48,7 +48,8 @@ void inserted_child_in_webelement_should_be_checked( final WebDriver driver ) th
 			fail( MISSING_ASSERTION_MSG );
 		} catch ( final AssertionError e ) {
 			Assertions.assertThat( e ).hasMessageContaining(
-					"option at 'html[1]/body[1]/form[3]/select[2]/option[5]':\n		was inserted" );
+					"html[1]/body[1]/form[3]/select[2]/option[5] at 'html[1]/body[1]/form[3]/select[2]/option[5]':\n"
+							+ "\t\twas inserted" );
```
diba1013: @roesslerj You seemed to have changed something with retest/recheck/#665. Is this expected?

beatngu13: Either we change `ElementDifferencePrinter` (and others affected) or we revert the PR.

roesslerj: Then I am in favor of changing the ElementDiferencePrinter...

beatngu13: There we have:

```java
private String createDescription( final ElementDifference difference ) {
	final IdentifyingAttributes attributes = difference.getIdentifyingAttributes();
	return attributes + " at '" + attributes.getPath() + "':";
}
```

How about:

```java
private String createDescription( final ElementDifference difference ) {
	final IdentifyingAttributes attributes = difference.getIdentifyingAttributes();
	final String toString = attributes.toString();
	final String path = attributes.getPath();
	return toString.equals( path ) //
			? "element at '" + path + "':" //
			: toString + " at '" + attributes.getPath() + "':";
}
```

roesslerj: How about  `return difference.getElement() + " at '" + attributes.getPath() + "':";`?

roesslerj: I think that would be the most correct (i.e. let the element decide how it's represented) and should now be the most readable.

beatngu13: OK good idea! 👍

roesslerj: I'll do it...

diba1013: `Element#toString` returns the retestId, which may not be readable (but solve the problem of getting the retestid). I think that we would want something like the type to be printed, because relying on the path can be quite ugly, since the path may be very long (e.g. Twitter tests). @lemoncurry realized this when changing the output with expected and actual for the cli that the console will break at the width, what makes the path unreadable because it is multiline.

> I think that would be the most correct (i.e. let the element decide how it's represented)

That was the problem on why I introduced the printers. The classes could not decide and there was a difference on how elements/differences are to be represented during debugging and how the user should perceive the output. Therefore I would suggest to have the printer decide a human readable way of displaying the element.

roesslerj: The current idea is that different Elements (e.g. Swing, HTML, XML, ...) have a different "best" representation. The retestId already addresses the problem of bringing the element down to the most simple and technology adapted _unique_ identification (e.g. for HTML use ID, name, tag + counter), that can even be customized. So I do think that using the retestId is the best way to go here.

beatngu13: Hmmm, I think retest ID + path is fine:

* retest ID can be used as a locator
* retest ID already tries to be readable and unique at the same time
* path can be used to lookup the element in the browser

diba1013:
@beatngu13 should we merge this so that we can test the rest?

PR #534, opened 2020-02-28
Feature/switch out fasterxml yaml

roesslerj:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Table of Contents
 * An exception is now thrown if explicit `Recheck#check` is called with an implicit `RecheckDriver`. This kind of mixing is not expected and therefore now prevented as it produced unexpected behavior due to both checks trying to find and create a Golden Master.
 * Improve the breaking messages from a `UnbreakableDriver` to include more information like the line number.
 * Add short cuts for screenshot enabling/disabling via the `RecheckWebOptions`.
+* Replace the dependency to fasterxml, which often caused trouble. Instead use snakeyaml.  
 
 
 --------------------------------------------------------------------------------
--- a/pom.xml
+++ b/pom.xml
@@ -54,7 +54,6 @@
 		<junit.jupiter.version>5.6.0</junit.jupiter.version>
 		<mockito.version>3.2.4</mockito.version>
 		<logback.version>1.2.3</logback.version>
-		<jackson.version>2.10.2</jackson.version>
 		<selenium.version>3.141.59</selenium.version>
 	</properties>
 
@@ -121,15 +120,9 @@
 		</dependency>
 
 		<dependency>
-			<groupId>com.fasterxml.jackson.dataformat</groupId>
-			<artifactId>jackson-dataformat-yaml</artifactId>
-			<version>${jackson.version}</version>
-		</dependency>
-
-		<dependency>
-			<groupId>com.fasterxml.jackson.core</groupId>
-			<artifactId>jackson-databind</artifactId>
-			<version>${jackson.version}</version>
+			<groupId>org.yaml</groupId>
+			<artifactId>snakeyaml</artifactId>
+			<version>1.21</version>            
 		</dependency>
 
 		<dependency>
--- a/src/main/java/de/retest/web/DefaultWebValueFinder.java
+++ b/src/main/java/de/retest/web/DefaultWebValueFinder.java
@@ -12,7 +12,14 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 
+import org.yaml.snakeyaml.DumperOptions;
+import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.Constructor;
+import org.yaml.snakeyaml.nodes.NodeId;
+import org.yaml.snakeyaml.nodes.Tag;
+import org.yaml.snakeyaml.representer.Representer;
+import org.yaml.snakeyaml.resolver.Resolver;
 
 import de.retest.recheck.ui.DefaultValueFinder;
 import de.retest.recheck.ui.descriptors.IdentifyingAttributes;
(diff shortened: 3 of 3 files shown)


beatngu13:
As mentioned in chat, I'm against switching from Jackson for the following reasons:

* Does what it should
* Well-known lib
* Code was simpler
* We use it in other projects for both YAML and JSON

But most importantly: it doesn't solve the actual problem. If a user uses a different version of SnakeYAML, the same thing can happen. You may "break" to project(s) of existing users. What are we going to do then, switch to another lib again?

IMHO the problem you described is due to an old Selenium version. From the feedback we get here, this isn't an issue elsewhere.

roesslerj:
Yaml is standard, so we are not bound to fasterXML. Well-known doesn't matter to me. It's already like the 5th time this causes problems, so we _have_ to address it. Code was actually much more ugly IMHO. 

The main reason fasterXML is a problem, is that it's shipped with Selenium. So we are _bound_ to run into those problems, whereas we don't have such a default problem with snakeyaml.

diba1013:
> IMHO the problem you described is due to an old Selenium version

Yes, that is true. But for some reason, users are not able to upgrade their Selenium version to the latest version.

A thing to note here. We only have one defaults.yml file that we want to read. Only for that reason, we have included/use the library. Therefore we are not really bound to this dependency and can exchange that or even remove that, if we decide to scrap the yml format&mdash;and use the filter syntax for example, which is totally capable of doing that.

beatngu13:
>Yaml is standard, so we are not bound to fasterXML.

Of course not, this is true.

>Well-known doesn't matter to me.

It should, as this means people know how to handle it.

>It's already like the 5th time this causes problems, so we _have_ to address it.

5 times? And all the other users that are fine? Here we have 179 stars and probably even more using it. You don't know how many are using SnakeYAML.

>Code was actually much more ugly IMHO.

Guess this is a matter of taste, but note that the amount of code doubled.

>The main reason fasterXML is a problem, is that it's shipped with Selenium. So we are _bound_ to run into those problems, whereas we don't have such a default problem with snakeyaml.

I couldn't find it in the dependency tree, can you point me to the trouble-causing dependency?

beatngu13:
> Yes, that is true. But for some reason, users are not able to upgrade their Selenium version to the latest version.

Thanks for pointing this out, but which user(s)? Where are the open issues? AFAIK we already ship with the newest Selenium version.

> A thing to note here. We only have one defaults.yml file that we want to read. Only for that reason, we have included/use the library. Therefore we are not really bound to this dependency and can exchange that or even remove that, if we decide to scrap the yml format—and use the filter syntax for example, which is totally capable of doing that.

Yep, that would be a _true_ fix, get rid of it entirely—if this is what we want. I really like the YAML-based format for maintaining the defaults, but we can move to something else. In the beginning we had plain text files, which we dropped because also the former `attributes.yml` grew in complexity.

diba1013:
I did some troubleshooting a while back for one or two users. However, I already deleted the emails to them... . 🙈 

Come to think of it, it could have been an issue with allure and not selenium per se.

However it seems that we need to discuss this a bit more and would suggest to not address this with this release.

beatngu13:
Thanks @diba1013!

If I see it correctly, we don't depend on Allure anymore? And current Selenium doesn't depend on Jackson?

As I said, I don't think switching to yet another lib is a good solution here. If we really want to get rid of the Jackson/YAML, I would suggest something like:

```
src/main/resources/defaults/
    |- all.txt
    |- a.txt
    |- address.txt
    ...
```

Where e.g. `all.txt`:

```
animation-delay="0s"
animation-duration="0s"
animation-iteration-count="1"
...
```

This shouldn't be too much work and we can easily handle this with plain Java.

roesslerj:
I don't really understand why we have such a lengthy discussion about it. YAML is a standard format. We can use any of like 5 libraries (https://yaml.org/ and fasterxml isn't even on the list). Switching it is minimal effort (30 lines change) and seems like minimal risk. We don't expose this to users, so well-known doesn't matter, and would in fact be a _counter_argument, making classpath issues more likely. Can we please just do this? We can even have both implementations and use a switch, depending on what we find on the classpath.

roesslerj:
And yes, we can have our own implementation, but this should be the _last_ option.

roesslerj:
> Here we have 179 stars and probably even more using it.

That IMHO is a very bad argument: There are people who don't have this problem, so I'm not gonna fix it?

roesslerj:
> but note that the amount of code doubled.

-25 +31 is now 6 lines more. That is hardly doubled. And the code is more readable, the amount of lines is due to our formatting with brackets and anonymous classes.

beatngu13:
>I don't really understand why we have such a lengthy discussion about it.

Because multiple people have a different opinion on how to approach this issue?

(Some) users are in dependency hell due to a lib we use internally. If we switch to yet another lib, this doesn't treat the cause but a sympton. Users that rely on SnakeYAML are possibly in the very same situation afterwards. Why is this "minimal risk"? This would just confront a different group of people (Jackson vs. SnakeYAML users) with the same problem.

This is why @diba1013 and I suggested to drop the YAML format and get rid of a corresponding dependency. This is also one of our overall goals—remove external dependencies—and is very common among libs, don't you agree?

>We don't expose this to users, so well-known doesn't matter, and would in fact be a _counter_argument, making classpath issues more likely.

I was referring to our devs, not users, which then don't have to use multiple tools for the same thing. (But I agree that UX has a higher priority than our laziness.)

On the other hand, if popularity is a counter-argument, why switch to something that is mentioned on yaml.org, whereas Jackson isn't?

>That IMHO is a very bad argument: There are people who don't have this problem, so I'm not gonna fix it?

This is not what I meant. I wanted to bring attention to a) the ratio of users that deals with this problem (because users have problems all the time) and b) the unknown amount of users that depends on SnakeYAML, for which we don't know the outcome.

>-25 +31 is now 6 lines more. That is hardly doubled. And the code is more readable, the amount of lines is due to our formatting with brackets and anonymous classes.

I was referring to the `readAttributesConfigFromFile` method, which grew a lot. But as I said, this is more a matter of taste and not my main point.

roesslerj:
Great, thanks!

PR #539, opened 2020-03-03
Improve RecheckWebOptions

beatngu13:


src/main/java/de/retest/web/RecheckWebOptions.java
```diff
@@ -36,22 +32,44 @@ public static RecheckWebOptionsBuilder builder() {
...
+		protected RecheckWebOptionsBuilder
 				checkNamingStrategy( final AutocheckingCheckNamingStrategy checkNamingStrategy ) {
 			this.checkNamingStrategy = checkNamingStrategy;
 			return this;
 		}
 
+		/**
+		 * @param screenshotProvider
+		 *            The {@link ScreenshotProvider} to be used. Default is determined via
+		 *            <code>retest.properties</code>.
```
diba1013: The default is&mdash;if nothing is specified in `retest.properties` and system properties&mdash;`ScreenshotProviders#DEFAULT`

src/main/java/de/retest/web/RecheckWebOptions.java
```diff
@@ -36,22 +32,44 @@ public static RecheckWebOptionsBuilder builder() {
...
+		 * @return self
+		 */
 		public RecheckWebOptionsBuilder screenshotProvider( final ScreenshotProvider screenshotProvider ) {
 			this.screenshotProvider = screenshotProvider;
 			return this;
 		}
 
+		/**
+		 * Disable screenshots via {@link ScreenshotProviders#NONE}.
```
diba1013: In some documentation, I often see:

> This is a shortcut method for `screenshotProvider( ScreenshotProviders.NONE )`

I would link to `screenshotProvider(ScreenshotProvider)`

src/main/java/de/retest/web/RecheckWebOptions.java
diba1013: Should we add that this overwrites the properties or can we assume that?

src/main/java/de/retest/web/screenshot/ViewportOnlyMinimalScreenshot.java
```diff
@@ -10,7 +10,8 @@
 import de.retest.recheck.ui.image.ImageUtils;
 
 /**
- * The default implementation that takes a screenshot of only the viewport visible to the user from the top page.
+ * The default implementation that takes a screenshot of only the viewport visible to the user from the top page. In
+ * contrast to {@link ViewportOnlyScreenshot}, images are resized to {@link #DEFAULT_WANTED_WIDTH_PX}.
```
diba1013:
```suggestion
 * The default implementation that takes a screenshot of only the viewport visible to the user from the top page. In
 * contrast to {@link ViewportOnlyScreenshot}, images are resized to maximum 800px if they exceed this width, while keeping the aspect ratio.
```

PR #543, opened 2020-03-04
Feature/naming strategy bugfixes

githubert:

- [ ] ~you updated necessary documentation within [retest/docs](https://github.com/retest/docs).~

## Description

Using this strategy with the `RecheckWebDriver` could end up throwing an `ArrayOutOfBoundsException` due to a missing check (if the parameter list ends up being empty). I also added a check to `enter`, because an empty list would not be of the expected type.

### State of this PR

All done.


beatngu13:
LGTM but you have to resolve the conflict on `CHANGELOG.md`.

PR #561, opened 2020-03-13
Feature/unify tab index

roesslerj:
- [x] the code follows the [Clean Code](https://clean-code-developer.com/) guidelines.
- [x] the necessary tests are either created or updated.
- [ ] all status checks (Travis CI, SonarCloud, etc.) pass.
- [x] your commit history is cleaned up.
- [x] you updated the changelog.
- [ ] you updated necessary documentation within [retest/docs](https://github.com/retest/docs).

PR #562, opened 2020-03-13
Feature/check clickability

roesslerj:
This is work in progress, but someone else has to take over...


beatngu13:
I think the issue of elements being outside of the current viewport, for which we cannot compute `isCovered`, should be addressed separately. This PR reduces the number of false positive that cause `ElementClickInterceptedException`, so I would consider this already a great improvement. What do you think?

varunnayak26:
I agree. Should there be a `TODO`?

beatngu13:
Yep, this is probably helpful. (And update the corresponding Jira ticket.)

roesslerj:
Also in favor of getting that in now and refine later (small steps).

varunnayak26:
I forgot this was marked as a Draft PR. Making it ready for review now.

src/test/resources/retest/recheck/de.retest.web.it.CenterIT/center-ChromeDriver.open.recheck/retest.xml
```diff
@@ -175,12 +187,16 @@
...
+									<key>covered</key>
+									<value xsi:type="xsd:string">false</value>
+								</entry>
 								<entry>
 									<key>shown</key>
 									<value xsi:type="xsd:string">true</value>
 								</entry>
 								<entry>
-									<key>tab-index</key>
+									<key>tabindex</key>
```
beatngu13: @roesslerj should there have been a migrator as part of #561? Also, why track when there is no value?

src/main/resources/javascript/getAllElementsByPath.js
```diff
@@ -281,6 +281,32 @@ function isDisabled(node) {
...
+	
+	    var topLeft = document.elementFromPoint(boundingLeft, boundingTop);
+	    var topRight = document.elementFromPoint(boundingRight, boundingTop);
+	    var bottomLeft = document.elementFromPoint(boundingLeft, boundingBottom);
+	    var bottomRight = document.elementFromPoint(boundingRight, boundingBottom);
+	    if ((topLeft != null && !node.contains(topLeft))
+	    	|| (topRight != null && !node.contains(topRight))
+	    	|| (bottomLeft != null && !node.contains(bottomLeft)) 
+	    	|| (bottomRight != null && !node.contains(bottomRight))) {
+	        return true;
```
beatngu13: @diba1013 do we use a specific JS formatter here? Looks a bit odd to me, and it seems like we are mixing tabs and whitespaces. I'd rather do:

```suggestion
	    if ((topLeft != null && !node.contains(topLeft))
	    	    || (topRight != null && !node.contains(topRight))
	    	    || (bottomLeft != null && !node.contains(bottomLeft)) 
	    	    || (bottomRight != null && !node.contains(bottomRight))) {
	        return true;
```

src/test/java/de/retest/web/it/SimplePageDiffIT.java
```diff
@@ -14,21 +15,23 @@
...
 	Recheck re;
 
-	@Before
+	@BeforeEach
 	public void setup() {
 		re = new RecheckImpl();
 		driver = WebDriverFactory.driver( Driver.CHROME );
 	}
 
+	@Disabled( "Unignore as soon as TODO from getAllElementsByPath.js is addressed." )
```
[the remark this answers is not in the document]
varunnayak26: I noticed while testing that simple-page-diff.html pushes an element(covered) outside the viewport which returns a false negative and results in an additional difference.

```
Test 'testSimpleChange' has 5 difference(s) in 1 state(s):
open resulted in:
	Metadata Differences:
	  Please note that these differences do not affect the result and are not included in the difference count.
		url:
		  expected="file:///home/varun/workspace/recheck-web/src/test/resources/pages/simple-page.html",
		    actual="file:///home/varun/workspace/recheck-web/src/test/resources/pages/simple-page-diff.html"
	div (twoblocks) at 'html[1]/body[1]/div[3]':
		id:
		  expected="twoblocks",
		    actual="changedblock"
	p (some_text) at 'html[1]/body[1]/div[3]/p[1]':
		text:
		  expected="Some text",
		    actual="Some changed text"
	p (some_more_text) at 'html[1]/body[1]/div[3]/p[2]':
		was deleted
	span (span-2) at 'html[1]/body[1]/div[5]/span[1]':
		covered:
		  expected="true",
		    actual="false"
	h2 (subheading) at 'html[1]/body[1]/h2[1]':
		was inserted">
```

src/main/resources/javascript/getAllElementsByPath.js
diba1013: I personally would have just used the vscode formatter, and we do have one for [recheck-web-js](https://github.com/retest/recheck-web-js/blob/master/.prettierrc.js) using [prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode). But I am afraid, that we did not setup any proper rules for that. 

Personally, I have gone with a similar to our existing java formatter, using a mix of tsfmt and tslint (soon to be replaced with eslint) to ensure some formatting rules in my typescript project.

src/test/resources/retest/recheck/de.retest.web.it.CenterIT/center-ChromeDriver.open.recheck/retest.xml
diba1013: We discussed that and agreed, that this is not necessary, as it is a fairly small change. Or to be more precise: The necessary work seems to big to add a migrator in recheck for recheck-web as we plan to include more breaking changes (this and pseudo-elements).

src/test/java/de/retest/web/it/IsCoveredIT.java
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	@Test
+	void partially_covered_element_should_be_clickable() {
+		driver.get( page( COVERED_PAGE ) );
+		final WebElement partiallyOverlappedElementOnRight =
+				driver.findElement( By.id( "partially-overlapped-element-right" ) );
+		final WebElement partiallyOverlappedElementOnLeft =
+				driver.findElement( By.id( "partially-overlapped-element-left" ) );
+		final Set<RootElement> rootElementOne = recheckAdapter.convert( partiallyOverlappedElementOnRight );
```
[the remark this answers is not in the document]
varunnayak26: I agree.

src/test/resources/retest/recheck/de.retest.web.it.CenterIT/center-ChromeDriver.open.recheck/retest.xml
beatngu13: OK, thx, this answers the first part. Regarding the no value: Do you know if this is common? If true, maybe we want to open issue in recheck for this?

src/main/resources/javascript/getAllElementsByPath.js
beatngu13: OK, since we hopefully soon integrate recheck-web-js, I mark this as resolved. We can have a separate clean-up PR.

src/test/java/de/retest/web/it/SimplePageDiffIT.java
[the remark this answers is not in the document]
beatngu13: But should we really disable the entire test because of this? Can we accept the change?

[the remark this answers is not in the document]
varunnayak26: I could update the message and the test will pass. However, as soon as the `TODO` is attended to, the test will fail again and we will need to revert back to the old message. What do you suggest is the better option?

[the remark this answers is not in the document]
beatngu13: I think enabling the test has more pros than cons. It simply shows the wrong computation of `covered` for elements outside of the viewport. But this is also the case in prod, so I assume it is fine. Would you agree?

src/test/resources/retest/recheck/de.retest.web.it.CenterIT/center-ChromeDriver.open.recheck/retest.xml
diba1013: I looked at the test and tried to reproduce this, as there is no apparent `tabIndex` specified. The code looks like it should work, defaults are properly specified and read.

The only way I could reproduce this is by `tabIndex=""`, but the test has no indication for that...

beatngu13: Hmmm, that's weird. We should look out for this in the future. Thanks a lot for testing it!

src/test/java/de/retest/web/it/SimplePageDiffIT.java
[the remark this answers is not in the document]
varunnayak26: From this point of view, it makes sense to enable the test.

PR #572, opened 2020-03-27
Run tests with newest JDK 14

beatngu13:


beatngu13:
First build failed, maybe needs a few more adaptions.

beatngu13:
I believe this is the issue:

https://community.sonarsource.com/t/issue-java-14-java-lang-noclassdeffounderror-java-util-jar-pack200/21853

I wonder why it works for recheck …

PR #581, opened 2020-04-17
Feature/autofocus

roesslerj:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Table of Contents
 ### Breaking Changes
 
 * Rename self-introduced property tab-index to tabindex, which is W3C standard.
+* Add `autofocus` as attribute (needs special handling).
 
 ### Bug Fixes
 
@@ -38,6 +39,7 @@ Table of Contents
 ### Improvements
 
 * Keep the order of the HTML elements preserved in the underlying result (internal change), which makes a couple things easier.
+
 
 --------------------------------------------------------------------------------
 
--- a/src/main/resources/javascript/getAllElementsByPath.js
+++ b/src/main/resources/javascript/getAllElementsByPath.js
@@ -282,6 +282,19 @@ function isDisabled(node) {
     return node.disabled ? true : false;
 }
 
+function hasAutofocus(node) {
+    if (!node.autofocus) {
+        return false;
+    }
+    if (node.autofocus === "") {
+        return false;
+    }
+    if (node.autofocus === "autofocus") {
+        return true;
+    }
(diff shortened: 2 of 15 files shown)


src/main/resources/javascript/getAllElementsByPath.js
```diff
@@ -282,6 +282,19 @@ function isDisabled(node) {
     return node.disabled ? true : false;
 }
 
+function hasAutofocus(node) {
```
diba1013, line 285: Just to drop in: @githubert What is the state of the recheck-web-js? Shouldn't this be done there`?

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ Table of Contents
 
 ### Improvements
 
+* Add `autofocus` as attribute (needs special handling).
```
diba1013, line 38: That would be a breaking change, wouldn't it?

src/main/resources/javascript/getAllElementsByPath.js
roesslerj: As long as this hasn't been finished, we need to keep them in sync. But from my understanding, this is a minimal change in the JS.

CHANGELOG.md
roesslerj: Iff an HTML contains autofocus, that change needs to be accepted. It surely isn't API breaking.

diba1013: > Breaking Changes when a change requires manual activity at update.

This is the case. While it is not an API change, it requires user activity. See the similarity to tabindex in the unrealeased section.

src/main/resources/javascript/getAllElementsByPath.js
diba1013: Then I would suggest that you create an corresponding PR in recheck-web-js? Just that this is not lost when transitioning. 

Just to clarify: I do not know what the best approach is here. In my understanding, this was already in progress. But let's wait what @githubert thinks.

githubert: An additional PR would be nice, so I can implement it over there. Once I figure out how to make the Internet Explorer happy, it'll be ready. (Last status was that it successfully runs the script and extracts the attributes, but then it goes wrong _somewhere_ in the whole process)

githubert: Or no PR, but open an issue.

roesslerj: Done: https://github.com/retest/recheck-web-js/issues/61

PR #582, opened 2020-04-17
Preserve the order of the elements in the HTML structure

roesslerj:

### DIFF
--- a/src/main/java/de/retest/web/mapping/PathsToWebDataMapping.java
+++ b/src/main/java/de/retest/web/mapping/PathsToWebDataMapping.java
@@ -3,9 +3,10 @@ package de.retest.web.mapping;
 import static org.apache.commons.lang3.StringUtils.countMatches;
 
 import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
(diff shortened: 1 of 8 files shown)


src/main/java/de/retest/web/mapping/PathsToWebDataMapping.java
```diff
@@ -36,13 +37,34 @@ public PathsToWebDataMapping( final Map<String, Map<String, Object>> mapping ) {
...
+		for ( final List<Object> entry : mapping ) {
+			if ( rootPath == null ) {
+				rootPath = entry.get( 0 ).toString();
+				continue;
+			}
+			if ( countMatches( entry.get( 0 ).toString(), "/" ) < countMatches( rootPath, "/" ) ) {
+				rootPath = entry.get( 0 ).toString();
+			}
+		}
+		return rootPath.replace( "//", "/" );
```
oezelenes, line 67: According to the sonarcloud check, we should check for null before replacing. 
This should do the job: `return rootPath != null ? rootPath.replace( "//", "/" ) : "";`

PR #588, opened 2020-05-08
Clean up travis config

beatngu13:


.travis.yml
```diff
@@ -37,7 +34,7 @@ deploy:
     on:
       tags: true
   - provider: releases
-    api_key: "${GH_TOKEN}"
```
diba1013, line 40: ~When looking at the [documentation](https://docs.travis-ci.com/user/deployment/releases/), the `api_key` (v1) is used. As far as I understood from basic skimming through, is `token` only usable with [version 2](https://docs.travis-ci.com/user/deployment-v2/) which is still in beta?~ Just noticed  that GitHub Releases part is considered stable.

Should we migrate the other options as well? E.g

1. `file_glob` is default true 
2. `skip_cleanup` has been replaced with `cleanup`? According to their blog, I would assume we do not have to specify that anymore...

diba1013, line 40: Maybe specify in the commit message that this uses the dpl v2 now.

beatngu13: I didn't want to switch to dpl v2, I thought this is also an alias in v1 (the info on Travis didn't imply this).

Will revert this commit, overall v2 is still in beta.

PR #598, opened 2020-06-03
Feature/fix close quits driver

diba1013:

- [x] ~you updated necessary documentation within [retest/docs](https://github.com/retest/docs).~ *Should be noted in several places to call the recheck lifecycle*

## Description

> TL;DR Do not mix the recheck lifecycle and driver lifecycle to prevent strange side-effects.

When experimenting with multiple open windows and tabs, I noticed that some strange behavior when closing some windows and tabs (via `driver.close()`)  and using the `RecheckDriver`, where in some instances `driver.close()` quit the whole driver instead of just closing the tab.

These issue solely appears when the `Recheck` instance is managed by an `AutocheckingRecheckDriver`.

### State of this PR

While I understand the intention of just using the `WebDriver` like you are used to (without calling specific recheck functions), the mixing of both lifecycles caused some weird behavior on side of recheck. 

I would therefore argue that you should use the extensions to handle the recheck lifecycle and call the driver lifecycle like you are used to. However, I am open for discussion (especially the `quit` method).

Since the documentation mentions on several pages to properly call the recheck lifecycle, I would not consider this a breaking change, because it should not hurt to call the lifecycle multiple times (as was the case when using a recheck extension).

### DIFF
--- a/src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
+++ b/src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
@@ -96,23 +96,6 @@ public class AutocheckingRecheckDriver extends UnbreakableDriver implements Rech
 	}
 
 	@Override
-	public void close() {
-		// Is this sensible? What about tests using separate sessions?
-		cap();
-		super.close();
-	}
-
-	@Override
-	public void quit() {
-		try {
-			// Is this sensible? What about tests using separate sessions?
-			cap();
-		} finally {
-			super.quit();
-		}
-	}
-
-	@Override
 	protected WebElement wrap( final WebElement element ) {
 		return AutocheckingWebElement.of( element, this );
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ Table of Contents
 
 ### Bug Fixes
 
+* Fix `AutocheckingRecheckDriver#close` and `AutocheckingRecheckDriver#quit` competing with the recheck lifecycle. This now requires that both lifecycles are called properly. Hint: Use the recheck extensions for the latter.
+
 ### New Features
 
 ### Improvements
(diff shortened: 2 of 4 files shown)


src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
```diff
@@ -96,23 +96,6 @@ public void get( final String url ) {
 		check( "get", null );
 	}
 
-	@Override
-	public void close() {
```
roesslerj, line 100: From the problem description I would assume that it suffices to remove the `close` method and keep the `quite` method...

diba1013: I should have mentioned this, but my thought was to have consistency. Since you can call `driver.close()` and the last window will quit the driver (according to documentation). But doing so does not call the `driver.quit()` method.

But as mentioned, I can do that, just wanted to share my thought process...

roesslerj: Hmm. Additional calls to `re.cap()` are not hurtful, right. So I would think we should have it with `driver.quit()`, even if not with `driver.close()`.

roesslerj: What do you think?

diba1013: After thinking about it, I would like to keep the proposed changes, i.e. separate both lifecycles. Since you still have to call other lifecycle methods (e.g. `startTest`, `capTest`; even though some are called transparently), why not also force users to call `cap` (assuming they do not want to run into the cap warner). 

Furthermore, we want users to use the test extensions, so that it manages the lifecycle for them (and we can transparently change some detail stuff in the background, e.g. call cap after all tests instead of just each test). 

Lastly, the `cap` method does IO (i.e. saving a report, uploading reports) which we do not want to do twice (especially uploading). But ultimately, we want to get rid of the `cap` method and upload only a changeset after each test.

PR #600, opened 2020-06-05
Perform a check when switching windows

diba1013:

- [x] you updated necessary documentation within [retest/docs](https://github.com/retest/docs). (*pending...*)

## Description

> TL;DR Switching windows or tabs now create a proper check.

Since switching windows introduces a completely new context (similar to a `WebDriver#get`), we want to perform a check for that.

Similar to the other `Autochecking...` implementations, the introduced `AutocheckingTargetLocator` will perform a check with an artificial name to be used by the `AutocheckingNamingStrategy` after the action in question has been performed, to allow for the new context to be captured.

### State of this PR

1. I am not sure, what other methods need a check, since it is important, in which context the script is executed. However, based on the `EventFiringTargetLocator` I assume that only the windows needs such checks, but I will need to do some testing for that.
1. I am assuming that the `TargetLocator` always returns its own driver and does not do any other modifications besides the method. All native implementations seem to do this.
1. ~I will add an integration (showcase) test for this feature later.~ Done

### DIFF
--- a/.retest/recheck.ignore
+++ b/.retest/recheck.ignore
@@ -41,6 +41,7 @@ matcher: type=meta, attribute-regex=.*
 
 # Have the tests pass locally:
 attribute=outline
+attribute=outline-width
 
 # So we can have a different ID in SimpleAutocheckingDriverShowcaseIT:
 matcher: xpath=html[1]/body[1]/form[1]/input[1], attribute=id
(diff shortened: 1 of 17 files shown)


diba1013:
While I have not done any tests for any additional methods&mdash;mostly because I struggle to come up with a proper use case. Thus I think that this is sufficient for a first draft.

diba1013:
I have fixed the test, since there were test failures due to the `covered` changes in #607.

PR #606, opened 2020-06-17
Update Golden Masters

diba1013:
Apparently, chrome updated once again... Hmm, just noticing. Shoudn't the metadata update correspondingly?


githubert:
Good question, I'd also expect the metadata to be updated.

diba1013:
> 
> 
> Good question, I'd also expect the metadata to be updated.

I will look into that one separately if you don't mind.

PR #612, opened 2020-07-06
Feature/retestid without gm

roesslerj:

## Description

TL;DR:
Allow the retestId to be used, even if the Golden Master is just being created. This is especially handy when the retestId is deterministic, and in conjunction with Surili.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Table of Contents
 ### Improvements
 
 * Always override the system metadata (i.e. on "architecture"), so that we do not end up with mixed info.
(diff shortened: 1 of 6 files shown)


roesslerj:
After consideration, I think this test should suffice as is :-)

diba1013:
> After consideration, I think this test should suffice as is :-)

Isn't that for me to decide 😄 ?

We had a discussion a while ago to what tests we consider as a showcase and what tests should test functionality. That means we have the showcase tests persist their Golden Masters within this repository, while the functional tests always assume no Golden Master created (i.e. temp directory). I would like to discuss this a bit more in depth with the team before continuing with this PR.

PR #613, opened 2020-07-06
Always override the system ones, so that we do not end up with mixed info

roesslerj:

## Description

TL;DR: We need to always override the system metadata, so that we do not mix it...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Table of Contents
 
 ### Improvements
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ Table of Contents
 
 ### Improvements
 
+* Always override the system metadata (i.e. on "architecture", so that we do not end up with mixed info.
```
diba1013, line 36: Missing closing parenthesis 😄

PR #628, opened 2020-07-22
Allow AutocheckingWebElement to implement Locatable

diba1013:

- [x] ~you updated necessary documentation within [retest/docs](https://github.com/retest/docs).~ *(not needed)*

## Description

> TL;DR This fixes unchecked casting within the Actions API.

This is related to RET-2071.

Apparently, the Actions API does some unchecked casting. When using any instance of `AutocheckingDriver` will cause this exception.

```
Actions builder = new Actions(driver);
builder.moveToElement(element).perform();
```

But, as the integration test shows, the actual method `Locatable#getCoordinates` is never called, but I added a sensible implementation and test for that. Since I need an actual browser for the implementation (because there is some weird stuff going on in the background), I made sure that no Golden Masters are created&mdash;to have the test pass and because they are not important in this context.

### State of this PR

The test failures are related to #627 and should be fixed once this PR is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ Table of Contents
 
 ### Bug Fixes
 
+* Fix unchecked `ClassCastExceptions` when using the `Actions`-API with the `AutocheckingDriver`.
+
 ### New Features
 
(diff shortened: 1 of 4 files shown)


diba1013:
@Mergifyio rebase

PR #630, opened 2020-07-30
Feature/element parent scope

roesslerj:

## Description

TL;DR Fix a bug which if the element is checked in isolation, some attributes can escape the scope.


roesslerj:
Please still review this, we can change/fix later. But need this now for customer.

PR #657, opened 2020-10-09
Add Option for autochecking delay

martin-v:

- [x] you updated necessary documentation within [retest/docs](https://github.com/retest/docs).

## Description

Add Option for autochecking delay

### State of this PR

finished

### DIFF
--- a/src/main/java/de/retest/web/RecheckWebOptions.java
+++ b/src/main/java/de/retest/web/RecheckWebOptions.java
@@ -19,12 +19,16 @@ public class RecheckWebOptions extends RecheckOptions {
 
 	private final AutocheckingCheckNamingStrategy checkNamingStrategy;
 	private final ScreenshotProvider screenshotProvider;
+	private final long autocheckingDelayMillis;
 
 	protected RecheckWebOptions( final RecheckOptions superOptions,
-			final AutocheckingCheckNamingStrategy checkNamingStrategy, final ScreenshotProvider screenshotProvider ) {
+			final AutocheckingCheckNamingStrategy checkNamingStrategy, final ScreenshotProvider screenshotProvider,
+			final long autocheckingDelayMillis ) {
 		super( superOptions );
 		this.checkNamingStrategy = checkNamingStrategy;
 		this.screenshotProvider = screenshotProvider;
+		this.autocheckingDelayMillis = autocheckingDelayMillis;
+
 	}
 
 	public static RecheckWebOptionsBuilder builder() {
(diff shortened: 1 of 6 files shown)


src/main/java/de/retest/web/RecheckWebOptions.java
```diff
@@ -19,12 +19,16 @@
 
 	private final AutocheckingCheckNamingStrategy checkNamingStrategy;
 	private final ScreenshotProvider screenshotProvider;
+	private final long autocheckingDelayMillis;
```
diba1013, line 22: Just a suggestion: In order for a more flexible unit, do we want to use `Duration`, so one could use `Duration.ofSeconds( 1 )`? Then we would not have to carry the unit with the variable.

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
```diff
@@ -35,7 +36,9 @@ public AutocheckingRecheckDriver( final RemoteWebDriver wrapped ) {
 	public AutocheckingRecheckDriver( final RemoteWebDriver wrapped, final RecheckOptions options ) {
 		super( wrapped );
 		this.options = options;
-		checkNamingStrategy = RecheckWebOptions.builder().build().getCheckNamingStrategy();
+		final RecheckWebOptions defaultOptions = RecheckWebOptions.builder().build();
```
diba1013, line 39: Just realizing, that we should propably do an `instanceof RecheckWebOptions` here, if someone used
```
RecheckOptions opts = RecheckWebOptions.builder()
    ...
    .setAutocheckingDelayMillis( 500 )
    .build()
WebDriver driver = new AutocheckingRecheckDriver(chrome, opts);
```
Like:
```suggestion
		final RecheckWebOptions defaultOptions = options instanceof RecheckWebOptions ? (RecheckWebOptions) options : RecheckWebOptions.builder().build();
```

diba1013 (changes requested):

src/main/java/de/retest/web/selenium/AutocheckingRecheckDriver.java
martin-v: You are right, I wasn't sure, but this failed:
```java
	private String t( final RecheckWebOptions o ) {
		return "web";
	}

	private String t( final RecheckOptions o ) {
		return "plain";
	}

	@Test
	void testName() throws Exception {
		final RecheckOptions o = RecheckWebOptions.builder().build();
		assertEquals( "web", t( o ) );
	}
```
I will make another commit to fix this.

src/main/java/de/retest/web/RecheckWebOptions.java
martin-v: I think the only sensible values for the delay are between 0 and 2000 millis. I guess with `Duration` it would be too easy to use too big values. 3000 sounds like a lot 3sec not. But in general a good idea.

diba1013 (approved):
Looks good. You may want to merge manually, if you do not want to deal with the minor build failures right now.

